PR #52, opened 2021-12-07
Replace cdn module with s3-public bucket

swiknaba:
Usage
```terraform
module "s3-frontend" {
  source = "github.com/dbl-works/terraform//s3-public?ref=v2021.11.13"

  # Required
  environment = "staging"
  project     = "someproject"
  bucket_name = "someproject-staging-frontend"

  # Optional
  versioning = false
  primary_storage_class_retention = 0
}
```


s3-public/README.md
```diff
@@ -0,0 +1,31 @@
...
+module "s3-cdn" {
+  source = "github.com/dbl-works/terraform//s3-public?ref=v2021.11.13"
+
+  # Required
+  environment = "staging"
+  project     = "someproject"
+  domain_name = "example.com"
+
+  # Optional
+  bucket_name_prefix = "cdn"
```
marcqualie, line 24: Can we not just pass in the bucket name instead? It's much more flexible, and we might not always want a bucket to include the domain.

In fact the domainis less important here, we actually probably want something like `cdn-{project}-{environment}" as the domain can change, or even have multiple domains per bucket (multi-tenancy).

marcqualie (approved):

PR #53, opened 2021-12-08
RDS now Creates an Initial DB

swiknaba:
setting up the initial DB is cumbersome﻿. This also helps to have the same naming convention everywhere.

https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/db_instance#name

> name - (Optional) The name of the database to create when the DB instance is created. If this parameter is not specified, no database is created in the DB instance. Note that this does not apply for Oracle or SQL Server engines. See the AWS documentation for more details on what applies for those engines. If you are providing an Oracle db name, it needs to be in all upper case.

:warning: this will force a re-creating of the RDS instance, hence we must ignore the `name` change.

- [x] Adjust the Notion documentation: we have to change the owner of the created database to be `_owner` rather than `root` (one for the roles to be created) -> @marcqualie .


rds/rds.tf
```diff
@@ -40,6 +44,7 @@ resource "aws_db_instance" "main" {
     ignore_changes = [
       engine_version, # AWS will auto-update minor version changes
       latest_restorable_time,
+      name, # if you didn't use this before it would re-created your RDS instance
```
marcqualie, line 47:
```suggestion
      name, # if you didn't use this before it would re-create your RDS instance
```

rds/rds.tf
```diff
@@ -15,9 +16,11 @@ resource "aws_db_instance" "main" {
   multi_az                            = var.multi_az
   publicly_accessible                 = var.publicly_accessible
   deletion_protection                 = true
+
   vpc_security_group_ids = [
     aws_security_group.db.id,
   ]
+
```
marcqualie, line 23: Why the extra spaces here? This looks really odd as there's no logical spearation of the attributes.

It makes sense for `tags` since we have them at the bottom of every resources, but thse are just generic key/value pairs.

rds/README.md
```diff
@@ -2,6 +2,8 @@
 
 Used for creating and configuring databases and their networking.
 
+Will create an initial database named `project_environment`.
```
marcqualie, line 5:
```suggestion
Will create an initial database named `{project}_{environment}`.
```

CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ This project does not follow SemVer, since modules are independent of each other
 
 
 ## Unreleased
+
+### RDS
+- an initial database is now created with the name `project_environment` [#52](https://github.com/dbl-works/terraform/pull/52)
```
marcqualie, line 13:
```suggestion
- an initial database is now created with the name `{project}_{environment}` [#52](https://github.com/dbl-works/terraform/pull/52)
```

marcqualie (approved):

rds/rds.tf
```diff
@@ -15,9 +16,11 @@ resource "aws_db_instance" "main" {
   multi_az                            = var.multi_az
   publicly_accessible                 = var.publicly_accessible
   deletion_protection                 = true
+
```
swiknaba, line 19:
```suggestion
```

rds/rds.tf
swiknaba, line 23:
```suggestion
```

swiknaba: removed.

PR #56, opened 2022-01-05
Block all public access to private buckets

marcqualie:
This also prevents accidently uploading public objects with an acl override.

Screen-Shot-2022-01-05-14-55-33 57


swiknaba (approved):
awesome

PR #59, opened 2022-01-10
Expose S3 usage permissions via group

marcqualie:
```terraform
resource "aws_iam_group_membership" "app-s3" {
  name = aws_iam_user.app.name
  group = module.s3-storage.group-usage-name
  users = [
    aws_iam_user.app.name
  ]
}
```

### DIFF
--- a/s3-private/iam.tf
+++ b/s3-private/iam.tf
@@ -4,7 +4,7 @@ resource "aws_iam_group" "usage" {
 
 resource "aws_iam_policy" "usage" {
   name        = "${var.project}-${var.environment}-s3-usage"
-  description = "For uploading/downloading encrypted storage files on S3"
+  description = "For uploading/downloading encrypted files on S3"
 
   policy = <<EOF
 {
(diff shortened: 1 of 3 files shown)


s3-private/iam.tf
```diff
@@ -0,0 +1,53 @@
...
+      "Effect": "Allow",
+      "Action": [
+        "s3:DeleteObject",
+        "s3:GetObject",
+        "s3:ListBucket",
+        "s3:PutObject",
+        "s3:PutObjectAcl"
+      ],
+      "Resource": [
+        "arn:aws:s3:::${var.project}-${var.environment}-storage",
```
swiknaba, line 31: Should probably use the `var.bucket_name`

s3-private/iam.tf
```diff
@@ -0,0 +1,53 @@
...
+      "Action": [
+        "s3:ListAllMyBuckets",
+        "s3:GetBucketLocation"
+      ],
+      "Resource": "arn:aws:s3:::*"
+    },
+    {
+      "Effect": "Allow",
+      "Action": [
+        "s3:DeleteObject",
```
swiknaba, line 24: Did we intentionally not add read/wrote to object versions?

marcqualie: Knew there was something missed off! We aren't currently using versions for the first example on uploads.

PR #62, opened 2022-01-31
Allow SQS access in ECS

UmarTayyab:


swiknaba (commented):
pls document the new variables in the Readme, see https://github.com/dbl-works/terraform/blob/main/ecs/README.md?plain=1#L32

then this is good to go :)

swiknaba (approved):

PR #68, opened 2022-04-13
[Chore] Replace S3 deprecated syntax

samkahchiin:
### Motivations
Lot of syntax in the S3 library is deprecated. (Check [here](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/s3_bucket))
Replace the deprecated syntax to ensure smoother transition to the latest version

## for testing
closes #70
closes #69

### DIFF
--- a/kms-key/kms-aliases.tf
+++ b/kms-key/kms-aliases.tf
@@ -1,4 +1,4 @@
 resource "aws_kms_alias" "a" {
-  name          = "alias/${var.project}/${var.environment}/${var.alias}"
+  name          = replace("alias/${var.project}/${var.environment}/${var.alias}", "/[^a-zA-Z0-9:///_-]+/", "-")
   target_key_id = aws_kms_key.key.key_id
 }
--- a/s3-public/s3.tf
+++ b/s3-public/s3.tf
@@ -1,24 +1,34 @@
 resource "aws_s3_bucket" "main" {
   bucket = var.bucket_name
-  acl    = "public-read"
-
-  versioning {
-    enabled = var.versioning
-  }
-
-  # Move data to a cheaper storage class after a period of time
-  lifecycle_rule {
-    enabled = var.primary_storage_class_retention == 0 ? false : false
-
-    noncurrent_version_transition {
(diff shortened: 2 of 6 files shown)


s3-public/s3.tf
```diff
@@ -1,24 +1,34 @@
...
+resource "aws_s3_bucket_lifecycle_configuration" "main-bucket-lifecycle-rule" {
+  bucket = aws_s3_bucket.main.id
 
+  rule {
+    id     = "primary-storage-class-retention"
+    status = var.primary_storage_class_retention == 0 ? "Disabled" : "Enabled"
     noncurrent_version_transition {
-      days          = var.primary_storage_class_retention == 0 ? 365 : var.primary_storage_class_retention
-      storage_class = "STANDARD_IA"
+      noncurrent_days = var.primary_storage_class_retention == 0 ? 365 : var.primary_storage_class_retention
```
samkahchiin, line 30: Params for noncurrent_version_transition

https://registry.terraform.io/providers/hashicorp/aws/3.75.0/docs/resources/s3_bucket_lifecycle_configuration#noncurrent_version_transition

kms-key/kms-aliases.tf
```diff
@@ -1,4 +1,4 @@
 resource "aws_kms_alias" "a" {
-  name          = "alias/${var.project}/${var.environment}/${var.alias}"
+  name          = replace("alias/${var.project}/${var.environment}/${var.alias}", "/[^a-zA-Z0-9:///_-]+/", "-")
```
swiknaba, line 2: Good! For old projects where the `replace` **might** actually be triggered, a change in the name will force a replacement.
Since we always refer to the ARNs using references through terraform modules (rather than pasting the literal ARN), and since we apply changes to an entire stack, not just a single resource, this should be fine.

We should still be careful when upgrading old projects though.

swiknaba (approved):
I've applied those changes to a test stack.

- Buckets are still in the correct state (especially their accessibility and encryption for the private bucket)
- the application (running in ECS) can still read and write files to both, the private and public bucket
- generating files for public access still works as expected without code changes
- files in the public bucket served via Cloudflare ("CDN setup") are still publicly accessible as expected

Great job! 👍 🎉

PR #71, opened 2022-05-02
Allow attaching custom policies to ECS

UmarTayyab:


swiknaba (approved):
Thanks! Let's also add `custom_policies` to the Readme under the optional variables, then we're good to merge

UmarTayyab:
> Thanks! Let's also add `custom_policies` to the Readme under the optional variables, then we're good to merge

Good catch, thanks! Done. 🙂

PR #72, opened 2022-05-13
[Chore] Stack module preparation

samkahchiin:
### Summary
- Add versions requirements to each of the modules

> Note: Only provider configurations are inherited by child modules, not provider source or version requirements. Each module must [declare its own provider requirements](https://www.terraform.io/language/providers/requirements). This is especially important for non-HashiCorp providers.

- Using minimum provider version instead of maximum provider version 

> Do not use ~> (or other maximum-version constraints) for modules you intend to reuse across many configurations, even if you know the module isn't compatible with certain newer versions. Doing so can sometimes prevent errors, but more often it forces users of the module to update many modules simultaneously when performing routine upgrades. Specify a minimum version, document any known incompatibilities, and let the root module manage the maximum version.

[Check here](https://www.terraform.io/language/providers/requirements#best-practices-for-provider-versions)

- Rename the file from `providers.tf` to `versions.tf` as suggested [here](https://www.terraform-best-practices.com/code-structure#getting-started-with-the-structuring-of-terraform-configurations)
- Set RDS username and password as `sensitive = true` so Terraform will then redact these values whenever they are used in the output of Terraform commands or log messages 

### Changelog
```
### Certificate, Cognito, ECR, ECS, Elasticache, KMS, NAT, RDS, S3-public, S3-private, Secret Manager, VPC, VPN
- Lock AWS provider to v4.x and terraform version to v1.x [#72](https://github.com/dbl-works/terraform/pull/72)
```


rds/rds.tf
```diff
@@ -6,7 +6,7 @@ resource "aws_db_instance" "main" {
   engine_version                      = var.engine_version
   instance_class                      = var.instance_class
   identifier                          = "${var.project}-${var.environment}"
-  name                                = "${var.project}_${var.environment}" # name of the initial database
+  db_name                             = "${var.project}_${var.environment}" # name of the initial database
```
swiknaba, line 9: 👍

elasticache/variables.tf
```diff
@@ -11,4 +11,13 @@ variable "node_count" {
   type    = number
   default = 1
 }
-# variable "engine_version" { default = "6.x" }
+
```
swiknaba, line 14: 👍 👍

rds/variables.tf
```diff
@@ -22,14 +22,21 @@ variable "multi_az" {
...
 # Credentials for the root RDS user
 # Only to be used in initial setup, never by applications
-variable "username" { default = "root" }
-variable "password" {}
+variable "username" {
+  default   = "root"
+  sensitive = true
+}
+variable "password" {
+  sensitive = true
```
swiknaba, line 30: 👍

s3-private/iam.tf
```diff
@@ -21,13 +21,11 @@ resource "aws_iam_policy" "usage" {
     {
       "Effect": "Allow",
       "Action": [
-        "s3:DeleteObject",
```
swiknaba, line 24: 👍 👍

elasticache/README.md
```diff
@@ -19,9 +18,11 @@ module "elasticache" {
   # optional
   node_count = 1
   node_type  = "cache.t3.micro"
+  engine_version = "6.x"
+  parameter_group_name = "default.redis6.x"
```
swiknaba, line 22: thanks for moving this to a variable!

vpn/README.md
```diff
@@ -75,24 +70,18 @@ copy paste the output into Outline Manager (formatted as JSON).
 
 Potentially, we can get this in Terraform, find a draft in the experimental PR [#33](https://github.com/dbl-works/terraform/pull/33).
 
-
```
swiknaba, line 78: we usually keep 3 new lines between headlines intentionally, to make it easier to quickly scan the readme when scrolling through it.

swiknaba (approved):
Great work! thank you for the thorough clean up! 🎉 

One nitpick: IMHO, it is easier to scan Readme files with 3 new lines between h1-headlines, especially if you just quickly skim some instructions to figure out if they contain what you are looking for.

PR #74, opened 2022-05-19
[Chore] Stack module

samkahchiin:
### Summary
A full stack module which will build a full environment at once, using all of our existing configurations and conventions.

###  Motivations
Starting new projects currently requires many setup steps and pulling together code from different repos, readmes, Notion docs. This creates an overhead of several days for setting up a new project.

### References
https://www.dropbox.com/scl/fi/w44zf1sdl99c2kd2ohv86/Mini-brief_-Terraform-%E2%80%9Cstack%E2%80%9D-module.paper?dl=0&oref=e&rlkey=hk6m55xt2in8lgz9qzql6933y


stack/README.md
```diff
@@ -0,0 +1,103 @@
...
+  elasticache_node_type  = "cache.t3.micro"
+
+  # vpc
+  vpc_cidr_block = "10.0.0.0/16"
+}
+```
+
+## NOTES
+
+1. To populate secrets to the secret manager, user need to prepare 2 json files
```
swiknaba, line 101: you will need the following secrets in the `${project}/app/${environment}` (or similar) vault:

* RAILS_MASTER_KEY
* database_url
* redis_url
* etc... 


just as a reminder, which secrets belong into which vault

stack/app/ecs.tf
```diff
@@ -0,0 +1,49 @@
...
+module "ecs" {
+  source = "../ecs"
+
+  project            = var.project
+  environment        = var.environment
+  vpc_id             = module.vpc.id
+  subnet_private_ids = module.vpc.subnet_private_ids
+  subnet_public_ids  = module.vpc.subnet_public_ids
+  secrets_arns = [
+    module.secrets["app"].arn
```
swiknaba, line 23: the secrets module now lives in a folder next to this one, i.e. it will live in its own terraform workspace, i.e. we cannot reference it via the `module` function here.

stack/setup/main.tf
```diff
@@ -0,0 +1,50 @@
...
+
+  # Optional
+  application = each.key
+  description = each.value.description
+}
+
+
+resource "aws_secretsmanager_secret_version" "app" {
+  secret_id     = module.secrets["app"].id
+  secret_string = file("${path.cwd}/app-secrets.json")
```
swiknaba, line 29: let's add `terraform-secrets.json` to a `.gitignore` file

swiknaba (commented):
```
│ Error: Error creating DB Instance: InvalidParameterCombination: Cannot find version 13.2 for postgres
│ 	status code: 400, request id: 5ef5c4f0-a084-42e7-8dbc-5a4f230283c3
│ 
│   with module.stack.module.rds.aws_db_instance.main,
│   on .terraform/modules/stack/rds/rds.tf line 1, in resource "aws_db_instance" "main":
│    1: resource "aws_db_instance" "main" {
```
-> copy-paste from the readme. Maybe we should just put the major version here.
Also: Can we please move the "optional" stuff to the very bottom? I stopped reading after I saw "optional" :D


image
image
image
my bad with the naming of the project! potentially we can strip/replace bad chars in the DB name with eg. underscore
image
same with some of the others (terraform has same string replace function)

image

maybe worth mentioning a minimum PW length here. For testing purposes, I was lazy and just typed "pw" in there.


other than that, all resources are created as expected 🎉 🚀 
image


will try to deploy a test application these days to verify nothing is missing 🙂 
Great work so far!!! 💪

stack/app/ecr.tf
```diff
@@ -0,0 +1,8 @@
+module "ecr" {
```
swiknaba, line 1: the ECR module is not environment aware, i.e. if we add 2 stacks for the same project, one stack per environment (e.g. "staging" and "production"), then we will have some collision here?

so far, we've opted to have 1 container registry per application, and use that for every environment.

WDYT @samkahchiin ?

stack/setup/main.tf
```diff
@@ -0,0 +1,50 @@
+locals {
```
swiknaba, line 1: what do you think about adding EIPs here? We'll need an EIP for the NAT (see https://stackoverflow.com/questions/61265108/aws-ecs-fargate-resourceinitializationerror-unable-to-pull-secrets-or-registry)

naming schema could be:

```
"${local.project}-${eip_name}"
```

and we create 1 EIP per `eip_name` given (as a list).
we'd also have to set `vpc` to true/false (true by default?)


```terraform
resource "aws_eip" "main" {
count = 3
```suggestion
locals {
```
  vpc = var.vpc
  tags = {
    Name        = "${local.project}-${eip_name}"
    Project     = local.project
    Environment = "production"
  }
}
```

File: stack/app/README.md [hunk 6]
```diff
@@ -0,0 +1,120 @@
...
+  # RDS
+  rds_instance_class     = "db.t3.micro"
+  rds_engine_version     = "13.2"
+  rds_allocated_storage  = 100
+
+  # ECS
+  allow_internal_traffic_to_ports = []
+  allowlisted_ssh_ips = []
+  grant_read_access_to_s3_arns = []
+  grant_write_access_to_s3_arns = []
```
swiknaba, line 90: @samkahchiin how can I access the `private_buckets_list`'s ARNs to pass them into ECS?

maybe by default, we should grant:
- write access to the private buckets (since those are meant for storing files form the app)
- read access to the private buckets (although write access does include read ATM)
- read access to the public bucket (e.g. to pull files from the CDN inside the app using AWS' internal APIs instead of performing a HTTPS call)

WDYT?

swiknaba: nvm sorry, you're already doing that! Should we mention this behaviour in a comment in the Readme maybe?

elasticache/variables.tf
```diff
@@ -3,7 +3,6 @@ variable "environment" {}
 variable "vpc_id" {}
 variable "vpc_cidr" {}
 variable "subnet_ids" { type = list(string) }
-variable "availability_zones" { type = list(string) }
```
swiknaba, line 6: we're using `availability_zones` in https://github.com/dbl-works/terraform/pull/77.

Let's not remove it here to avoid merge conflicts :)

PR #76, opened 2022-05-27
[Chore] IAM Management

samkahchiin:
### Summary
Using Attribute Based Access Control (ABAC) to assign the permissions based on project and environment tags.

Tags | Values
-- | --
Environment | staging, production
Project | eg. youtube, google (separated by `:`)

Example of assigning the principal tags to the User/Group
```
  tags = {
      staging-developer-access-projects    = "youtube:google"
      production-developer-access-projects = "youtube"
  }
```  

### Motivations
AWS has a hard limit on the number of policies that can be attached to a IAM user 
- 10 groups/user
- 20 policies/user

Refer to [here](https://aws.amazon.com/premiumsupport/knowledge-center/iam-increase-policy-size/). In some projects, we already are hitting these limits and thus are blocked in our regular flow.

### What should you expect
- After the implementation, user should be able to

Tag | What user can do
-- | --
`<env>-developer-access-projects` | User should be able to carry out `List, Describe, and Get` actions in the relevant projects<sup>1</sup> in `<env>` 
`<env>-admin-access-projects` | User should be able to carry out all actions<sup>2</sup> of the relevant projects in `<env>` 

- <sup>1</sup> Some of the action does not support resource tag, thus actions should still be able to be carried out by User regardless of projects principal tag (eg. `ListAllMyBuckets`)
- <sup>2</sup> User should not be able to access specific module, eg. IAM management, billing 

### References
- The policy simulator doesn't work well with the Principal Tag. (ResourceTag is working fine though 🤷‍♀️ )
eg. Even if the user's principal tag value match with the resource tag value, it won't display 'allowed' in the policy simulator
https://policysim.aws.amazon.com/home/index.jsp
- Changes in the policy sometimes can take more than 10 mins to take effects
https://stackoverflow.com/questions/20156043/how-long-should-i-wait-after-applying-an-aws-iam-policy-before-it-is-valid#:~:text=In%20most%20cases%2C%20changes%20you,time%20expire%20and%20be%20replaced.
- List of actions of each resources 
https://docs.aws.amazon.com/service-authorization/latest/reference/reference_policies_actions-resources-contextkeys.html

#### Naming Convention

> Use _ (underscore) instead of - (dash) everywhere (resource names, data source names, variable names, outputs, etc).

> Beware that actual cloud resources often have restrictions in allowed names. Some resources, for example, can't contain dashes, some must be camel-cased. The conventions in this book refer to Terraform names themselves.

https://www.terraform-best-practices.com/naming

### Related PR
https://github.com/lh-innovationhub/infrastructure/pull/177


samkahchiin:
@swiknaba  
This PR would allow the production admin to have access to everything related to the project including the production DB.
However, in some project such as flightvoucher, we would not want the admin to have access to the prod DB as well, an example will be the roles of Matjaz in the flightvoucher project.

Would you think the cases of flightvoucher is an exception cases?
Or it is quite common in all of our existing project

swiknaba:
> 

I'd say having DB access included by default is OK. Especially for new projects, that is anyway needed during the set-up phase. Tougher restrictions can always be added later manually.

swiknaba (commented):
could you please add CI for the new `iam-policy-for-toggable-resources`? see https://github.com/dbl-works/terraform/pull/87 for reference. Thanks!

PR #77, opened 2022-06-02
[Chore] Use elasticache replication group instead of cluster

samkahchiin:
### Summary

In the terraform documentations, it is stated that the value for `num_cache_nodes` for Redis must be 1, which prohibited us for creating more redis cache nodes.
This PR uses replication group which is the AWS [recommended way](https://docs.aws.amazon.com/AmazonElastiCache/latest/red-ug/Replication.CreatingRepGroup.html) to manage replica nodes of Redis 

Please be noted that this setup disable cluster mode.

### DIFF
--- a/elasticache/elasticache.tf
+++ b/elasticache/elasticache.tf
@@ -23,6 +23,10 @@ resource "aws_elasticache_replication_group" "main" {
     "${aws_security_group.main.id}",
   ]
 
+  # Cluster Mode
+  num_node_groups         = var.num_node_groups
+  replicas_per_node_group = var.replicas_per_node_group
+
   tags = {
     Project     = var.project
     Environment = var.environment
--- a/elasticache/elasticache.tf
+++ b/elasticache/elasticache.tf
@@ -10,6 +10,7 @@ resource "aws_elasticache_replication_group" "main" {
   subnet_group_name          = aws_elasticache_subnet_group.main.name
   at_rest_encryption_enabled = true
   kms_key_id                 = var.kms_key_arn
+  snapshot_retention_limit   = var.snapshot_retention_limit
   # When you change an attribute, such as engine_version,
   # by default the ElastiCache API applies it in the next maintenance window.
   # Because of this, Terraform may report a difference in its planning phase
--- a/elasticache/elasticache.tf
+++ b/elasticache/elasticache.tf
@@ -1,13 +1,15 @@
 resource "aws_elasticache_replication_group" "main" {
-  replication_group_id = "${var.project}-${var.environment}"
-  description          = "Collection of Redis cache clusters for ${var.project}-${var.environment}"
-  engine               = "redis"
-  node_type            = var.node_type
-  num_cache_clusters   = var.node_count
-  parameter_group_name = var.parameter_group_name
-  engine_version       = var.engine_version
-  port                 = 6379
-  subnet_group_name    = aws_elasticache_subnet_group.main.name
+  replication_group_id       = "${var.project}-${var.environment}"
+  description                = "Collection of Redis cache clusters for ${var.project}-${var.environment}"
+  engine                     = "redis"
+  node_type                  = var.node_type
+  num_cache_clusters         = var.node_count
+  parameter_group_name       = var.parameter_group_name
+  engine_version             = var.engine_version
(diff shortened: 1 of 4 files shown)


samkahchiin:
@swiknaba should we also have a setup for cluster mode as well?

Factors to consider when choosing between Redis (cluster mode disabled) or Redis (cluster mode enabled) 
https://docs.aws.amazon.com/AmazonElastiCache/latest/red-ug/Replication.Redis-RedisCluster.html

swiknaba:
> @swiknaba should we also have a setup for cluster mode as well?
> 
> Factors to consider when choosing between Redis (cluster mode disabled) or Redis (cluster mode enabled) https://docs.aws.amazon.com/AmazonElastiCache/latest/red-ug/Replication.Redis-RedisCluster.html

@samkahchiin 

if I understand the terraform docs right, to enable cluster-mode, all we have to do is append `.cluster.on` to the `parameter_group_name` and set

```
  num_node_groups         = 2
  replicas_per_node_group = 1
```
?

so potentially, we can have these two vars optional? Or do you think it might be cleaner, to write a second script for that?

elasticache/elasticache.tf
```diff
@@ -1,15 +1,23 @@
...
+  replication_group_id       = "${var.project}-${var.environment}"
+  description                = "Collection of Redis cache clusters for ${var.project}-${var.environment}"
+  engine                     = "redis"
+  node_type                  = var.node_type
+  num_cache_clusters         = var.node_count
+  parameter_group_name       = var.parameter_group_name
+  engine_version             = var.engine_version
+  port                       = 6379
+  subnet_group_name          = aws_elasticache_subnet_group.main.name
+  at_rest_encryption_enabled = true
```
swiknaba, line 11: great addtition!

let's also add `snapshot_retention_limit` and default it to `0`. Have it as a variable and mention in the readme that 0 is no backups. We need backups for some apps.

samkahchiin:
@swiknaba correct! 
I think we can have it in the same module and let the user decide whether they want to use cluster mode by setting the value.

elasticache/variables.tf
```diff
@@ -21,3 +22,18 @@ variable "engine_version" {
...
   default = "6.x"
 }
+
+variable "snapshot_retention_limit" {
+  type    = number
+  default = 0
+}
+
+variable "replicas_per_node_group" {
+  default     = ""
```
swiknaba, line 32:
```suggestion
  default     = 0
```

```

│ Error: Incorrect attribute value type
│ 
│   on .terraform/modules/stack/elasticache/elasticache.tf line 28, in resource "aws_elasticache_replication_group" "main":
│   28:   replicas_per_node_group = var.replicas_per_node_group
│     ├────────────────
│     │ var.replicas_per_node_group is ""
│ 
│ Inappropriate value for attribute "replicas_per_node_group": a number is required.
```

swiknaba, line 32:
```suggestion
  default     = 0
```

```

│ Error: Incorrect attribute value type
│ 
│   on .terraform/modules/stack/elasticache/elasticache.tf line 28, in resource "aws_elasticache_replication_group" "main":
│   28:   replicas_per_node_group = var.replicas_per_node_group
│     ├────────────────
│     │ var.replicas_per_node_group is ""
│ 
│ Inappropriate value for attribute "replicas_per_node_group": a number is required.
```

elasticache/variables.tf
```diff
@@ -21,3 +22,18 @@ variable "engine_version" {
   type    = string
   default = "6.x"
 }
+
+variable "snapshot_retention_limit" {
+  type    = number
+  default = 0
+}
+
+variable "replicas_per_node_group" {
```
swiknaba, line 31: let's also add that to the variables in the stack module

swiknaba, line 31: let's also add that to the variables in the stack module

swiknaba: https://github.com/dbl-works/terraform/pull/86/commits/56eaed9766df5325d71c292952862ae151799590

swiknaba: https://github.com/dbl-works/terraform/pull/86/commits/56eaed9766df5325d71c292952862ae151799590

PR #78, opened 2022-06-09
Execution role is based on region

marcqualie:
When spreading out of more than a single region the IAM role name needs to be unique.

By default, this will be backwards compatible. I would prefer for it to be regional by default, but this would require updating all previous projects using this.


swiknaba (approved):
Great thanks! Can we mention the new variable in the readme of the ECS module as well?

PR #89, opened 2022-06-17
[Chore] Improve Stack module

samkahchiin:
### Summary
Solve the following issues

1. Setting AZ for Redis Cluster Mode will throw error (`When specifying preferred availability zones,  the number of cache clusters must be specified and must match the number of preferred availability zones`)
- https://github.com/hashicorp/terraform-provider-aws/issues/5104
> For Redis Cluster Mode Enabled replication groups (e.g. when using cluster_mode in Terraform), we currently do not have the ability to set availability zones via the NodeGroupConfiguration parameter, which would likely require changing the cluster_mode argument.

- As suggested [here](https://stackoverflow.com/questions/59194562/terraform-elasticache-redis-cluster-with-specified-availability-zones)
> if you specify VPC subnets for your replication group via subnet_group_name key, cache instances will be created in the AZs of those subnets

which has already been covered by our existing codes

_TLDR: We can enable multi-AZ but can't specify AZ for cluster mode now using TF_ 

2.Outputs of the stack module are missing:
- target groups arn (ecs + ssh) => for the ecsx deploy config
- ~secrets arn => for the ecsx deploy config~
_NOTE: This value should be retrieved in the stack-setup output and passed in as variables in the stack module_
- security group arn => for the ecsx deploy config
- subnets (private + public) => for the ecsx deploy config

3. Setup certificate creation + validations in the stack setup instead of the cloudflare module
4. Set the elasticache group in the stack module as cluster mode.


swiknaba (approved):
great thanks!

PR #90, opened 2022-06-21
Stack Module Fixes

swiknaba:
- remove superfluous secrets
- use existing certificate module to avoid duplication

### DIFF
--- a/stack/setup/main.tf
+++ b/stack/setup/main.tf
@@ -70,12 +70,12 @@ resource "aws_acm_certificate_validation" "default" {
 
 ## EIP
 resource "aws_eip" "nat" {
-  for_each = var.eips_nat_count
+  count = var.eips_nat_count # 1 NAT/IP per region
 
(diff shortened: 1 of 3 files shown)


stack/setup/main.tf
```diff
@@ -49,18 +49,15 @@ module "secrets-kms-key" {
   deletion_window_in_days = var.kms_deletion_window_in_days
 }
 
-resource "aws_acm_certificate" "main" {
-  domain_name = var.domain
-
-  subject_alternative_names = var.add_wildcard_subdomains ? ["*.${var.domain}"] : []
+module "aws-acm-certificate" {
```
samkahchiin, line 52: This will causes error in setup/cloudflare.tf

swiknaba: moved to https://github.com/dbl-works/terraform/pull/91 because I cannot reference a branch that contains a `/`

PR #95, opened 2022-06-29
[Chore] Add ECS iam management

samkahchiin:
### Summary
Module for ECS IAM policy.


iam/iam-policy-for-ecs/main.tf
```diff
@@ -0,0 +1,352 @@
...
+    local.staging_read_access_projects,
+    local.production_read_access_projects
+  )
+
+  write_access_projects = concat(
+    local.staging_write_access_projects,
+    local.production_write_access_projects
+  )
+
+  region = "eu-central-1"
```
samkahchiin, line 35: Hardcoded this value for now but will think about how to manage this dynamically.
Perhaps we can take the assignment of policy to the user from the module.

swiknaba: @marcqualie should we prepend the region to the policy names, that we need for each geo region?

iam/iam-policy-for-ecs/taggable-resources/variables.tf
```diff
@@ -0,0 +1,17 @@
+variable "region" {
```
swiknaba, line 1: can we add a README to this? :-)

iam/iam-policy-for-ecs/core/main.tf
```diff
@@ -0,0 +1,103 @@
+data "aws_iam_user" "main" {
+  user_name = var.username
+}
+
+locals {
+  staging_developer_access_projects    = try(data.aws_iam_user.main.tags.staging-developer-access-projects, "")
```
swiknaba, line 6: is it possible to keep this general, i.e. not bound to environments named `staging` and `production`?

samkahchiin: Can do if we pass the projects and environment as variables.
The tradeoff is we can't automatically assign the permission based on the user tags using the modules
(We have to extract the values and passing them as variables into the modules)
Both have its pros and cons
Option 1: Extracting the tag values inside module -> Simplified the usage of module, i.e. just need to pass the username into the module, less flexibility
 Option 2: Making the variable more general -> More flexibility, need to extract the values of projects name and environment from tag (if relevant) and pass into the modules

swiknaba (approved):
Tested on our DBL Test project. works 👍 

Thanks 🎉

PR #97, opened 2022-07-01
RDS as read replica, Stack with VPC Peering

swiknaba:
Adds the possibility to launch a stack with an RDS read replica and VPC Peering to connect to a "main"/"master" stack.

- new module: VPC Peering
- RDS can now be launched as read-replica
- stack can now be launched with a RDS read replica and vpc-peering
- require aws config alias for stack
- fix passing multiple providers to submodule
- Update README.md
- CI hack for terraform bug

## testing
Members of DBL can see the terraform scripts for launching a 2nd stack in another geo region here: https://github.com/dbl-works/test-application/pull/6

Screenshot 2022-07-03 at 20 29 35

Screenshot 2022-07-03 at 20 28 08

### DIFF
--- a/stack/app/rds.tf
+++ b/stack/app/rds.tf
@@ -33,4 +33,6 @@ module "rds" {
   multi_az               = var.environment == "production"
   master_db_instance_arn = var.rds_master_db_instance_arn
   is_read_replica        = var.rds_is_read_replica
+  regional               = var.regional
+  name                   = var.rds_name
 }
--- a/stack/app/rds.tf
+++ b/stack/app/rds.tf
@@ -1,5 +1,6 @@
 module "rds-kms-key" {
   source = "../../kms-key"
+  count  = var.rds_master_db_kms_key_arn == null ? 1 : 0
 
   # Required
   environment = var.environment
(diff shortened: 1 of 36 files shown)


stack/app/rds.tf
```diff
@@ -9,6 +10,7 @@ module "rds-kms-key" {
 
   # Optional
   deletion_window_in_days = var.kms_deletion_window_in_days
+  multi_region            = var.rds_multi_region_kms_key
```
samkahchiin, line 13: What happened if this is a rds replica, and no `var.rds_master_db_kms_key_arn` is provided? 

Should we do some sort of check to make sure user is using the replica key 

```
multi_region = var.rds_is_read_replica ? false : var.rds_multi_region_kms_key
```
or is it okay to have a non replica key as read-replica rds's kms key?

samkahchiin (approved):
LGTM except some minor feedbacks and questions. 
GJ 🎉 👏

stack/app/rds.tf
swiknaba: good point. The read-replica must use the master KMS key, unless the master DB does not use encryption at rest.

swiknaba:
> LGTM except some minor feedbacks and questions. GJ 🎉 👏

Thanks for the review! Added all the new outputs, those make definitely sense. Here the outputs of the sample application:

image

image

PR #101, opened 2022-07-14
[Chore] Add Cross Region Replication for S3

samkahchiin:
### Summary
- Add Cross Region Replication for S3
- Add S3 module which consists of the most basic s3 setup

### Notes
- Terraform disallow us to declare provider in the module and we can only pass declared provider to the module.
`Providers cannot be configured within modules using count, for_each or depends_on.`
- AWS allows bi-directional replicating

## ⚠️ On Upgrade

Due to a new abstraction layer on the bucket modules, we see the following change:
```
# before
module.storage-bucket.aws_s3_bucket.main

# after
module.storage-bucket.module.s3.aws_s3_bucket.main
```

Hence perform the upgrade as follows:

```shell
terraform state rm module.storage-bucket.aws_s3_bucket.main
terraform init -upgrade
terraform import module.storage-bucket.module.s3.aws_s3_bucket.main "YOUR-BUCKET-NAME"
```


swiknaba (commented):
LGTM so far! Thank you!

samkahchiin:
@swiknaba  I have added some changes. Can you have a look again on this? 😄

s3-private/s3.tf
```diff
@@ -1,15 +1,16 @@
-resource "aws_s3_bucket" "main" {
-  bucket = var.bucket_name
-
-  tags = {
-    Name        = var.bucket_name
-    Project     = var.project
-    Environment = var.environment
-  }
+module "s3" {
```
swiknaba, line 1: we have to be careful when using this new module, since moving the "main" reference into a module will force a re-creation.

before: `module.storage-bucket.aws_s3_bucket.main`
after    : `module.storage-bucket.module.s3.aws_s3_bucket.main`

samkahchiin:
Some Add-ons:
1. AWS has remove the `replicate existing object` options in the replication rules therefore we now can only use s3 batch replication which is not supported by terraform. 
Therefore, after setting up the replica, we need to do the one time batch job to replicate existing object to replica.
-> https://aws.amazon.com/blogs/aws/new-replicate-existing-objects-with-amazon-s3-batch-replication/

2. Enable RTC (Make sure the obj is copied within 15 mins) will be charged extra with the following fees:
```
S3 Replication Time Control data transfer† | $0.015 per GB
```
RTC will also automatically turn on replication metrics which create 3 metrics per replication rules (Monitor the total number of S3 API operations that are pending replication, the total size of objects pending replication, and the maximum replication time to the destination Region)

The charges are similar to Cloudwatch
~ 0.30 * 3 = 0.90 per month per RTC replication rules
https://aws.amazon.com/cloudwatch/pricing/

samkahchiin:
@swiknaba  Some updates of my findings:
- Replication can't be triggered by another CRR.
i.e.

| US | EU | AP |
|----|----|----|
| Replicate to EU | Replicate to US & AP | Replicate to EU |
 
- When we upload an object to US, it will only trigger replication in the EU bucket
- CRR rules in EU won't be triggered by replication, therefore, it won't trigger a replication to US bucket

Therefore,
if we want to trigger multi directional replication, we need to explicitly stated the CRR rules we would like to have, eg.
| US | EU | AP |
|----|----|----|
| Replicate to EU **& AP** | Replicate to US & AP | Replicate to EU **& US** |

swiknaba (approved):
Great job, thank you! Fantastic to have some graphs in the readme to visualize concepts.

I had to add a few more fixes to the VPC-Peering part to test this with the stack module in a multi-region test deploy.

I've run all these changes against our multi-region test deployment; seems to work 🙂

PR #103, opened 2022-07-25
Allow launching stack with VPC peering without RDS read-replica

swiknaba:
* Allow launching stack with VPC peering without RDS read-replica
* "bug" fix: if passed duplicated IP (ranges) for the ecs-ssh ingress policy, we get a hard error. We now ensure, this is a unique list

### DIFF
--- a/ecs/security-groups.tf
+++ b/ecs/security-groups.tf
@@ -62,7 +62,7 @@ resource "aws_security_group" "ecs" {
 resource "aws_security_group_rule" "ecs" {
   for_each                 = toset(var.allow_internal_traffic_to_ports)
   type                     = "ingress"
-  from_port                = each.key ## ??
+  from_port                = each.key
   to_port                  = each.key
   protocol                 = "tcp"
   security_group_id        = aws_security_group.ecs.id
--- a/ecs/security-groups.tf
+++ b/ecs/security-groups.tf
@@ -62,7 +62,7 @@ resource "aws_security_group" "ecs" {
 resource "aws_security_group_rule" "ecs" {
   for_each                 = toset(var.allow_internal_traffic_to_ports)
   type                     = "ingress"
-  from_port                = 0
+  from_port                = each.key ## ??
   to_port                  = each.key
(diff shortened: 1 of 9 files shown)


stack/app/ecs.tf
```diff
@@ -28,10 +28,10 @@ module "ecs" {
 
   allow_internal_traffic_to_ports = var.allow_internal_traffic_to_ports
 
-  allowlisted_ssh_ips = flatten(concat([
+  allowlisted_ssh_ips = distinct(flatten(concat([
```
swiknaba, line 31: fixes

```terraform
│ Error: Error authorizing security group rule type ingress: InvalidParameterValue: The same permission must not appear multiple times
│ 	status code: 400, request id: XXX
│ 
│   with module.stack.module.ecs.aws_security_group_rule.ecs-ssh[0],
│   on .terraform/modules/stack/ecs/security-groups.tf line 82, in resource "aws_security_group_rule" "ecs-ssh":
│   82: resource "aws_security_group_rule" "ecs-ssh" {
```

samkahchiin (approved):
LGTM except the security group rule part that I am not sure. 👍

ecs/security-groups.tf
```diff
@@ -62,7 +62,7 @@ resource "aws_security_group" "ecs" {
 resource "aws_security_group_rule" "ecs" {
   for_each                 = toset(var.allow_internal_traffic_to_ports)
   type                     = "ingress"
-  from_port                = 0
```
samkahchiin, line 65: I don't quite understand about this security group rule, so not really sure whether this changes is valid.

swiknaba: we open up the port for the web app by default. If we want to use any other port for some service that is not our web server, e.g. a pdf-renderer running in its own service, then we need to allow traffic for this port.

E.g. in `allow_internal_traffic_to_ports` we would pass `[ 5000 ]` to allow traffic for port 5000. In the current setting, we would now open up all ports from 0 - 5000. This was never intended.

PR #105, opened 2022-08-02
Allow passing multiple CIDR blocks to Elasticache Security Group Rule

swiknaba:
Optionally allow that we can access a Redis cluster in another VPC/region.

Tested on the test application ✅


samkahchiin (approved):
LGTM

PR #109, opened 2022-08-13
Elasticache: Allow opt-out of automatic failover

marcqualie:
Smaller/test projects don't need the failover. Requires a minimum of 2 nodes to be active which doubles the cost.


swiknaba (approved):
Makes sense, thank you!

PR #110, opened 2022-08-16
[SC 2864] Add maxmemory_policy to elasticache

samkahchiin:
## Summary
Allow user to pass in `maxmemory_policy` to elasticache. If it is not passed, default value will be set as `volatile-lru`.

### DIFF
--- a/elasticache/elasticache.tf
+++ b/elasticache/elasticache.tf
@@ -86,6 +86,9 @@ resource "aws_elasticache_parameter_group" "main" {
 
   name = local.cluster_name
   # Family need to be specified so that we can copy keys n values from the existing parameter group
+  # The regexp will extract the family name from the parameter group name, eg:
+  # default.redis6.x => redis6.x
+  # default.redis6.x.cluster.on => redis6.x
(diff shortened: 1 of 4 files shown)


elasticache/elasticache.tf
```diff
@@ -80,3 +80,21 @@ resource "aws_elasticache_replication_group" "cluster_mode" {
     ]
   }
 }
+
+resource "aws_elasticache_parameter_group" "main" {
+  count = var.maxmemory_policy == null ? 0 : 1
+
+  name = local.cluster_name
+  # Family need to be specified so that we can copy keys n values from the existing parameter group
```
swiknaba, line 88: can we document a few examples of which part we return, e.g. when "parameter_group_name" is
* `default.redis6.x` => `redis6.x`
* `default.redis6.x.cluster.on` => `redis6.x`

swiknaba (approved):

elasticache/elasticache.tf
samkahchiin: Good ideas! Have added it into the comments :)

PR #116, opened 2022-08-22
cloudwatch: Observability alerts for cloudwatch module

samkahchiin:
### Summary
We should have a centralized dashboard to track all of the important metrics for:
1. ECS Container
2. RDS
3. Redis

It will also send alarms to Slack when the metrics exceed the threshold


stack/app/README.md
```diff
@@ -128,6 +128,7 @@ module "stack" {
   cloudwatch_cluster_names        = []
   # https://aws.amazon.com/rds/instance-types/
   db_instance_class_memory_in_gb  = 1
+  db_is_read_replica              = true
```
swiknaba, line 131: we also have `rds_is_read_replica` which is a bit confusing. Should we either use the existing `rds_is_read_replica`, or prefix it with the module that needs this variable, i.e. `cloudwatch_is_db_read_replica` ?

samkahchiin: Agree. I'll use the existing `rds_is_read_replica`

PR #117, opened 2022-08-25
xray: add xray module and grant xray permissions to ecs and stack 

samkahchiin:
#### Summary
Add xray module and grant xray permissions to ECS and stack


stack/app/README.md
```diff
@@ -114,6 +114,8 @@ module "stack" {
   secret_arns                     = []
   # appends region to the name (usually ${project}-${environment}) for globally unique names
   regional                        = true
+  enable_xray                     = true
```
swiknaba, line 117:
```suggestion
  enable_xray                     = false
```

should we put the actual default value here?

swiknaba (approved):
Great thanks!

xray/xray-sampling.tf
```diff
@@ -0,0 +1,18 @@
+# Refer to here for the sampling rules
```
swiknaba, line 1: what do we need this for?

swiknaba (approved):
LGTM!

xray/xray-sampling.tf
samkahchiin: Do u mean the comments or the files?

samkahchiin: Do you refer to the comments? 
Leave a note here so we have a docs reference of the sampling rules. Not a really important notes actually

swiknaba: ah, so this is excluding traces for the health check endpoints, good thinking!

PR #123, opened 2022-09-08
[Features] Setting up Cloudwatch -> Fivetrans -> Snowflake pipeline

samkahchiin:
#### Summary
Syncing the following Cloudwatch metrics to Snowflake table thru Fivetran:
- CPUUtilization
- MemoryUtilization
- TargetResponseTime (p50, p90, p99)
- 5xx Error Count


cloudwatch-snowflake/README.md
```diff
@@ -0,0 +1,40 @@
...
+  fivetran_group_id = "fivetran-group-id"
+  organisation = "meta"
+  fivetran_api_key = XXXXXXXXXXX
+  fivetran_api_secret = YYYYYYYYYY
+  tracked_resources_data = [
+    {
+      projectName      = "facebook"
+      serviceName      = "web",
+      clusterName      = "facebook-production"
+      loadBalancerName = "app/facebook-production/7ed111aa2bf1de37"
```
rubyconvict, line 23: - how do we obtain the value?

cloudwatch-snowflake/main.tf
```diff
@@ -0,0 +1,78 @@
...
+    name = "LambdaInvokePolicy"
+
+    policy = jsonencode({
+      Version = "2012-10-17"
+      Statement = [
+        {
+          Sid      = "InvokePermission"
+          Effect   = "Allow"
+          Action   = ["lambda:InvokeFunction"]
+          Resource = "*"
```
rubyconvict, line 67: - couldn't/shouldn't this be precise?

cloudwatch-snowflake/tracker/utils/cloudwatch.js
```diff
@@ -0,0 +1,159 @@
...
+  const metricDataQueries = resourcesData.flatMap((data) => performanceMetricQueries(data))
+
+  const prevHour = dateUtil.previousHour()
+  const params = {
+    MetricDataQueries: metricDataQueries,
+    StartTime: dateUtil.oneHourBefore(prevHour),
+    EndTime: prevHour
+  }
+
+  // NOTE: A single GetMetricData call can include as many as 500 MetricDataQuery structures.
```
rubyconvict, line 136: - why 500?

cloudwatch-snowflake/variables.tf
```diff
@@ -0,0 +1,35 @@
+variable "fivetran_group_id" {
+  type        = string
+  description = "Also know as external_id. Understand the group concept here: https://fivetran.com/docs/getting-started/powered-by-fivetran#creategroupusingtheui"
```
rubyconvict, line 3: this link anchor is currently dead, it is missing _a_ in: 'create**~a~**group'
should be:
https://fivetran.com/docs/getting-started/powered-by-fivetran#createagroupusingtheui

cloudwatch-snowflake/tracker/utils/cloudwatch.js
```diff
@@ -0,0 +1,159 @@
...
+  })
+}
+const getCloudwatchData = async () => {
+  const resourcesData = JSON.parse(constants.RESOURCES_DATA)
+  const metricDataQueries = resourcesData.flatMap((data) => performanceMetricQueries(data))
+
+  const prevHour = dateUtil.previousHour()
+  const params = {
+    MetricDataQueries: metricDataQueries,
+    StartTime: dateUtil.oneHourBefore(prevHour),
```
rubyconvict, line 132: I admire the simplicity of this solution, but I thought we were generally interested in importing historical data, first, to have a lot of data already. Could you double-check with Fried, in case that was his expectation?

cloudwatch-snowflake/tracker/utils/cloudwatch.js
samkahchiin: 500 is the number of metrics data that we can receive in 1 API call. It is stated in the [AWS docs](https://docs.aws.amazon.com/AmazonCloudWatch/latest/APIReference/API_MetricDataQuery.html)

cloudwatch-snowflake/variables.tf
samkahchiin: Thanks!

cloudwatch-snowflake/main.tf
samkahchiin: @rubyconvict  
Good catch. 
This could be precise but I think it shouldn't be, because
- This role is created for the purpose of connecting fivetran-lambda
- If we make it too specific (i.e. limited the access to only 1 specific lambda function), we would have to create 1 role for each lambda function that we create in the future (i.e. we will then have a lot of lambda role in the AWS IAM)
- I have made it an optional creation, so if user has created the lambda role somewhere, they can pass the lambda arn and skip the creation.

What do you think?

cloudwatch-snowflake/README.md
samkahchiin: This is the `aws_lb => arn_suffix`. 
Have updated the comment :D

cloudwatch-snowflake/main.tf
rubyconvict: Ok, I'm no expert at all, but not using '*' seems to be a popular opinion in general, so I had to mention it.
BTW, do we practice use of https://github.com/tenable/terrascan or anything like that check security concerns for us (at least running it locally, if not on CI)? CC: @swiknaba

cloudwatch-snowflake/tracker/utils/cloudwatch.js
rubyconvict: taking metrics from last day, we can get 5metrics*24h=120structures, right?

cloudwatch-snowflake/main.tf
swiknaba: @rubyconvict do you want to write a short proposal for it? https://go/propose
Unsure how much sense it makes to run such scans on these modules, since many modules are flexible enough to use them securely or less securely. So probably these scans should be part of the CI pipeline of project-specific terraform scripts? -> if so, we can put that into our base CI script here: https://github.com/dbl-works/monorepo-scaffold ?

cloudwatch-snowflake/tracker/utils/cloudwatch.js
samkahchiin: @swiknaba. 
Yes, we can write a script to export the data based on our needs.
Different project might have different needs (i.e. different time range) so the script can be vary.

PR #125, opened 2022-09-16
Add DB instructions to connect DB to Fivetran

MatjazKavcic:


rds/README.md
```diff
@@ -117,3 +117,121 @@ function generate_rds_password () {
...
+35.235.32.144/29
+```
+#### SSH tunneling
+
+If the application DB is behind bastion, an SSH tunnel has to be configured to allow access:
+
+1. Bastion URL is the same you use to access the database.
+2. Please read how to setup bastion with public key [here](https://github.com/dbl-works/terraform/pull/118/files#diff-94cb490f08aaef66dedfd461d69ed21f92555c728a81212e41accf096b31c118)
+
+Check [Modifly connector](https://fivetran.com/integrations/postgres_rds/?name=modifly&groupId=plausible_comprise) for example.
```
swiknaba, line 237: let's write some generic example here

MatjazKavcic: I don;t have experience, can you add an example @swiknaba?

PR #130, opened 2022-09-30
cloudwatch-snowflake: Update lambda function name

samkahchiin:
#### Summary
Update lambda function name and add lambda arn in the outputs so it can be reused


swiknaba (approved):
💪

PR #133, opened 2022-09-30
allow passing path to custom lambda & improve query id regex

swiknaba:
In case we need to deploy different lambdas across multiple regions, this change allows us to do so

Also, if the stats is `Maximum` or `p99.9` the current regex to clean the Query Id name was insuffcient.

image

### DIFF
--- a/cloudwatch-snowflake/tracker/utils/cloudwatch.js
+++ b/cloudwatch-snowflake/tracker/utils/cloudwatch.js
@@ -14,8 +14,9 @@ const PERCENTILES = [
   'p50'
 ]
 
+// expected pattern: ^[a-z][a-zA-Z0-9_]*$
 const formatQueryId = (name) => {
-  return name.replace(/-|\//g, '_')
+  return name.replace(/-|\/|\./g, '_')
 }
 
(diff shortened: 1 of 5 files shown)


samkahchiin (approved):
LGTM 👍

cloudwatch-snowflake/tracker/utils/cloudwatch.js
```diff
@@ -46,7 +47,7 @@ const ecsMetricQueries = ({ serviceName, clusterName, projectName, environment }
 
 const responseTimesQueries = ({ projectName, loadBalancerName, environment }) => {
   return PERCENTILES.map(percentile => ({
-    Id: `${percentile}_${formatQueryId(loadBalancerName)}_${projectName}_${environment}`, /* /^[a-z][a-zA-Z0-9_]*$./ */
+    Id: `${formatQueryId(percentile)}_${formatQueryId(loadBalancerName)}_${projectName}_${environment}`, /* /^[a-z][a-zA-Z0-9_]*$./ */
```
swiknaba, line 50: Fixes issues when the percentile includes a dot, e.g. `p99.9`

PR #135, opened 2022-09-30
Adds SQL Script For A Readonly User/Role

swiknaba:
Adds a SQL script to create a readonly role/user on Snowflake to be used e.g. in Tableau/Metabase/Grafana


samkahchiin (approved):
LGTM 👍

PR #136, opened 2022-10-03
Autoscaling/ECS: Setup New Module

samkahchiin:
#### Summary
Add autoscaling policy for ECS. 
This module should allow the ECS to autoscale based on the metric threshold + send notification to Slack when it scale in/out.

#### Some thoughts
1. Why Step Scaling is chosen instead of Target Tracking
- Target Tracking only limit us to trigger the scaling with limited ECS Metrics
image
 - Target tracking doesn't give us the flexibility to set the scale down threshold (It automatically create the scale down alarm for us)

2. Why don't we use individual metric alarm?
If we use individual metric alarm, the task count will be increased too quick (> 1) when metrics increase.
Eg. When both memory and cpu utilization is beyond the threshold, it will increase by 2
However, the autoscaling would not be triggered if only 1 metric > threshold.
Eg. When memory is beyond the upper threshold, but cpu is < step down threshold, the increment will be +1 - 1 = 0
And no task count will be added.

3. Why don't we use composite alarm instead of using the metric alarm?
Composite alarm doesn't allow us to use autoscaling as the alarm actions


samkahchiin:
@swiknaba  
I have removed the role from the module :)

swiknaba (approved):
approving this, since we can create a new IAM module for the autoscaling role in a new PR

swiknaba (approved):
👍 🚀 ❤️

samkahchiin:
> Can we move the IAM role to here: https://github.com/dbl-works/terraform/tree/main/iam ? Then we have it defined in a single place, which makes updating the role simpler across projects

I forgot about this! 🙈 
Have added an ecs autoscaling IAM module

PR #137, opened 2022-10-07
fivetran/connectors/lambda: Improve and extract lambda module

samkahchiin:
#### Summary
Abstract the lambda further to allow us to pass in any lambda to the Fivetran module to create any number of connectors using lambdas


morozRed:
https://app.shortcut.com/squake/story/4055/update-terraform-infra-make-fivetran-connector-a-bit-more-generic-so-it-can-deploy-any-lambda-with-passed-params-e-g-redis

fivetran/connectors/lambda/README.md
```diff
@@ -0,0 +1,101 @@
+# Terraform Module: Fivetran lambda
```
swiknaba, line 1: Should we add a folder with example lambdas, where we can put e.g. the cloudwatch lambda?

fivetran/connectors.tf
```diff
@@ -54,3 +52,26 @@ resource "fivetran_connector" "github" {
     pat          = each.value.pat
   }
 }
+
+module "lambda_connector" {
+  for_each = { for lambda in var.sources_lambda : join("-", compact([lambda.service_name, lambda.project, lambda.environment])) => lambda }
+  source   = "../fivetran/connectors/lambda"
+
+  providers = { fivetran = fivetran }
```
swiknaba, line 60: Do we also have to pass in a AWS provider if we want to deploy lambdas into multiple regions? E.g. see http://go/infra-fivetran-multi-region-lambda

samkahchiin: How about just passing the AWS_REGION as env variables?
https://github.com/dbl-works/terraform/pull/137/files#diff-b6ed8ac68eaaeb7c15e9437b253a9c87749872d3f796d438e53703f69d0f2e02R91

swiknaba: if that works to deploy to multiple regions, sure

fivetran/README.md
```diff
@@ -28,6 +28,45 @@ module "fivetran" {
...
   destination_database_name    = "${project}-${environment}"
+
+  # Sources
+  ## Github
+  sources_github = [
+    {
+      organisation = "facebook"
+      sync_mode    = "SpecificRepositories"
+      repositories = ["facebook/react"]
+      username     = "squake-bot"
```
swiknaba, line 39:
```suggestion
      username     = "dbl-bot"
```

fivetran/connectors.tf
samkahchiin: I was thinking that we can passing the AWS_REGION as an env variables so the script can fetch the cloudwatch metrics from the region we set. However, it doesn't seems to be a good idea now because
1. AWS_REGION is an reserved env variables and cannot be overwritten.
```
InvalidParameterValueException: Lambda was unable to configure your environment variables because the environment variables you have provided contains reserved keys that are currently not supported for modification. Reserved keys used in this request: AWS_REGION
```
2. Even I manage to overwrite it by changing the env name, it also seems weird that a lambda function in US region can fetch data from EU cloudwatch (?)

iam/iam-policy-for-fivetran-lambda/README.md
```diff
@@ -0,0 +1,12 @@
+## IAM for fivetran lambda
+
+Creates a IAM role which connect lambda and fivetran
```
swiknaba, line 3:
```suggestion
Creates an IAM role that connects Lambda and Fivetran.
```

fivetran/connectors.tf
```diff
@@ -54,3 +52,31 @@ resource "fivetran_connector" "github" {
     pat          = each.value.pat
   }
 }
+
+module "lambda_connector" {
+  for_each = { for lambda in var.sources_lambda : join("-", compact([lambda.service_name, lambda.project, lambda.environment])) => lambda }
```
swiknaba, line 57: Will the combination of service/project/environment result in duplicated values for a multi-regions setup? If so, should the region code be included in the unique key?

fivetran/connectors.tf
```diff
@@ -54,3 +52,31 @@ resource "fivetran_connector" "github" {
...
+  source   = "../fivetran/connectors/lambda"
```
swiknaba, line 58:
```suggestion
  source   = "./connectors/lambda"
```

would this short-cut work as well?

fivetran/connectors/lambda/fivetran.tf
```diff
@@ -0,0 +1,18 @@
+resource "fivetran_connector" "lambda" {
```
swiknaba, line 1: Will this resource be created regardless of having `sources_lambda` defined?

fivetran/connectors/lambda/fivetran.tf
```diff
@@ -0,0 +1,18 @@
...
+  run_setup_tests   = true
+
+  destination_schema {
+    name = try(var.connector_name, local.function_name)
+  }
+
+  config {
+    function = aws_lambda_function.main.function_name
+    role_arn = var.lambda_role_arn
+    region   = var.aws_region_code
```
swiknaba, line 16: https://registry.terraform.io/providers/fivetran/fivetran/latest/docs/resources/connector#region

The documentation mentions `region` and `aws_region_code` ... is there any benefit of using one over the other?

fivetran/connectors.tf
```diff
@@ -54,3 +52,31 @@ resource "fivetran_connector" "github" {
...
+  # optional
+  service_name       = each.value.service_name
+  aws_region_code    = each.value.aws_region_code          # lambda's aws region
+  lambda_role_arn    = var.lambda_settings.lambda_role_arn # Lambda role created for connecting the fivetran and lambda. Reuse the same role if you already have it created.
+  lambda_source_dir  = each.value.lambda_source_dir
+  lambda_output_path = each.value.lambda_output_path
+  script_env         = each.value.script_env
+}
+
+resource "aws_iam_role_policy_attachment" "fivetran_policy_for_lambda" {
```
swiknaba, line 78: Great idea!

fivetran/connectors/lambda/fivetran.tf
samkahchiin: I think it is just the different name that fivetran use to define the input variables

eg.
aws_region_code:
https://fivetran.com/docs/rest-api/connectors/config#dynamodb

region:
https://fivetran.com/docs/rest-api/connectors/config#amazonads

swiknaba (commented):
Great work so far! Let's test this once e2e, if all goes as expected I think we can merge this 🚀

fivetran/connectors/lambda/fivetran.tf
samkahchiin: If sources_lambda is not defined, the default value will be `[]` and nothing will be created

samkahchiin:
@swiknaba 

Regarding the creation of multi region lambda, it is unfortunately not doable (as far as I can think of) in fivetran module because:
1. It is not encouraged to have provider in the child module (https://github.com/hashicorp/terraform/issues/22907#issuecomment-535580358)
```
if you put a provider configuration inside a child module then removing that module will fail any time there are resources present using that provider in the module, because by removing the module you've also removed the provider configuration Terraform would need in order to destroy those objects.
``` 
as it will threw error when we remove the module

2. We can do it by passing the provider in the provider field, but unfortunately we can't instantiating the provider in a loop (looping thru the lambda_source variables)
=> https://github.com/hashicorp/terraform/issues/19932

3. Even we created the provider in the root module, we can't define the provider aliases with string interpolation
https://discuss.hashicorp.com/t/defining-provider-aliases-with-string-interpolation-not-working-in-terraform-0-12/1569

Therefore, the last resort is to ask the user to create the lambda function from other region on their own. :(
=> Added an example in README.md (https://github.com/dbl-works/terraform/pull/137/files#diff-94cb490f08aaef66dedfd461d69ed21f92555c728a81212e41accf096b31c118R78)

fivetran/connectors.tf
samkahchiin: Refer to this comment -> https://github.com/dbl-works/terraform/pull/137#issuecomment-1281859748

Since we only can create lambda in one region, I think this is OK?

swiknaba (approved):

PR #140, opened 2022-10-20
autoscaling/ecs: Support target response time as ECS autoscaling metrics

samkahchiin:
#### Summary
Support target response time as ECS autoscaling metrics

#### Motivation
ECS can be scaled up based on the target response time and therefore enhance the user experience time


swiknaba (approved):
Great addition, thank you!

PR #145, opened 2022-10-28
fivetran/connectors/google_analytics: Add google analytics fivetran module

samkahchiin:
#### Summary
Add fivetran google analytics module

### DIFF
--- a/fivetran/connectors/google_analytics/README.md
+++ b/fivetran/connectors/google_analytics/README.md
@@ -1,0 +1,64 @@
+# Terraform Module: Fivetran Google Analytics
+
+### NOTES
+If auth credentials are not passed in, you need to log in to the Fivetran UI and manually authorize a Google Account to sync data from Google Analytics
+
+### Example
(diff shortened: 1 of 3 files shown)


swiknaba (dismissed):
LGTM! Have we tested this module somwhere?

samkahchiin:
Yes.
Tested here in https://github.com/lh-innovationhub/lih-datastack/pull/3/files
I can't test the auth because we need to setup a new user and someone has to give it access for that.
The consequence for this is, it will override the previous user whenever this module is applied, thus have to re-authorize the apps again. Other than that, it looks fine

samkahchiin:
Yes.
Tested here in `go/ga_fivetran`
I can't test the auth because we need to setup a new user and someone has to give access to it.
The consequence for this is, it will override the previous user whenever this module is applied, thus have to re-authorize the apps again. Other than that, it looks fine

swiknaba (approved):

PR #146, opened 2022-10-28
Allow Creating an ECS Cluster without Secrets/KMS Keys

swiknaba:
#### Summary

Some simple apps might not need any secrets, thus this allows to launch an ECS Cluster without passing secret ARN(s) and KMS key ARN(s)

#### Motivation

I'm toying with an app that does not need secrets, and I don't want to launch resources that won't be needed.

### DIFF
--- a/ecs/execution-role.tf
+++ b/ecs/execution-role.tf
@@ -167,7 +167,7 @@ resource "aws_iam_role_policy" "ecs-task-execution-secrets-policy" {
           "kms:Decrypt",
           "kms:GenerateDataKey" # for writing to an encrypted S3 bucket
         ],
-        "Resource" : kms_and_secret_arns,
+        "Resource" : length(local.kms_and_secret_arns) > 0 ? kms_and_secret_arns : "can-t-be-blank",
       }
     ]
   }) : []
--- a/ecs/execution-role.tf
+++ b/ecs/execution-role.tf
@@ -153,6 +153,10 @@ locals {
   ])
 }
 
+# Making secret/kms ARNs optional was added later.
+# To avoid a breaking change by introducing a `count` logic (the name would change),
+# we added a dummy policy without any effect instead. AWS performs some validations
+# on policies, hence we can't just use a blank string.
 resource "aws_iam_role_policy" "ecs-task-execution-secrets-policy" {
   name = "ecs-task-execution-secrets-policy"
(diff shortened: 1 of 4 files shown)


ecs/execution-role.tf
```diff
@@ -146,11 +146,18 @@ resource "aws_iam_role_policy" "ecs-task-execution-policy" {
...
+    var.kms_key_arns,
+  ])
+}
+
 resource "aws_iam_role_policy" "ecs-task-execution-secrets-policy" {
   name = "ecs-task-execution-secrets-policy"
   role = aws_iam_role.ecs-task-execution.name
 
-  policy = jsonencode({
+  policy = length(local.kms_and_secret_arns) > 0 ? jsonencode({
```
samkahchiin, line 160: I think we have to use count instead of making the policy an empty string because it will throw error below:
```
"policy" contains an invalid JSON policy
```

samkahchiin (changes requested):

ecs/execution-role.tf
swiknaba: I wanted to avoid count, since that would change the resource name in the terraform state, and thus it can break existing implementations :(

ecs/execution-role.tf
```diff
@@ -160,10 +167,18 @@ resource "aws_iam_role_policy" "ecs-task-execution-secrets-policy" {
...
+    ]
+    }) : jsonencode({
+    "Version" : "2012-10-17",
+    "Statement" : [
+      {
+        "Effect" : "Deny",
+        "Action" : [
+          "secretsmanager:GetSecretValue",
+        ],
+        "Resource" : "arn:aws:secretsmanager:${var.region}:*:secret:can-t-be-blank",
```
samkahchiin, line 181: How about just skip the creation of this role policy if kms and secret arns is missing?

samkahchiin (approved):
LGTM. 👍  😄

ecs/execution-role.tf
samkahchiin: Can we add a comment to explain why we have the dummy policy?

swiknaba: done! :)

samkahchiin (approved):
Nice. Thanks! 🙇‍♀️

PR #147, opened 2022-11-07
cloudwatch: Add p99 and average to CPU and Memory metrics graph

samkahchiin:
#### Summary
Add p99 and average to CPU and Memory metrics graph

### DIFF
--- a/cloudwatch/main.tf
+++ b/cloudwatch/main.tf
@@ -143,7 +143,7 @@ locals {
             ["...", { "id" : "m3", "label" : "p99", "stat" : "p99" }],
           ],
           "view" : "timeSeries",
-          "stacked" : true,
+          "stacked" : false,
           "region" : "${var.region}",
           "title" : "Memory Usage",
           "period" : var.metric_period,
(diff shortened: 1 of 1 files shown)


cloudwatch/main.tf
```diff
@@ -138,14 +138,15 @@ locals {
         "type" : "metric",
         "properties" : {
           "metrics" : [
-            ["AWS/ECS", "MemoryUtilization", "ServiceName", "web", "ClusterName", name, { "label" : name }]
+            ["AWS/ECS", "MemoryUtilization", "ServiceName", "web", "ClusterName", name, { "id" : "m1", "label" : "Max", "stat" : "Maximum" }],
+            ["...", { "id" : "m2", "label" : "Avg", "stat" : "Average" }],
+            ["...", { "id" : "m3", "label" : "p99", "stat" : "p99" }],
           ],
           "view" : "timeSeries",
           "stacked" : true,
```
swiknaba, line 146: let's set `stacked` to false, stacking these makes no sense:

See e.g. this example, average and max value are both at about 50%, but it seems the average is at 100... 

image

swiknaba (approved):
p99 for ECS is WIP for AWS, we can leave this in though since it is expected to show numbers at some point. wdyt?

PR #148, opened 2022-12-06
[Chore] Add output to stack app

samkahchiin:


swiknaba (approved):
👍 with the stack we must have an NLB, hence this output should work.

PR #151, opened 2023-01-01
Support multiple SSL Certificates per ALB (SNI)

swiknaba:
#### Summary

- Adds Basic SNI Support
- also allow stack to pass additional certificates through

#### Motivation

https://aws.amazon.com/blogs/aws/new-application-load-balancer-sni/
https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/lb_listener_certificate

Given a multi-region application, I want to add certificates for
* api.my-app.test
* api.eu-central-1.my-app.test
* api.us-east-1.my-app.test

to test a specific region directly. This is not doable with wildcard certificates due to the usage third level domains.

#### Test plan

Tested here:

https://go/terraform-151-1 # api.my-app.test/healthz
https://go/terraform-151-2 # api.eu-central-1.my-app.test/healthz
https://go/terraform-151-3 # api.us-east-1.my-app.test/healthz

**before**
image

### DIFF
--- a/ecs/alb.tf
+++ b/ecs/alb.tf
@@ -39,4 +39,12 @@ resource "aws_alb_listener" "https" {
     target_group_arn = aws_alb_target_group.ecs.arn
     type             = "forward"
   }
+
+  # Using SNI to attach multiple certificates to the same load balancer
+  resource "aws_alb_listener_certificate" "https" {
+    for_each = var.additional_certificate_arns
+
+    listener_arn    = aws_alb_listener.https.arn
+    certificate_arn = each.value
+  }
 }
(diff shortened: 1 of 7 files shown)


ecs/alb.tf
```diff
@@ -40,3 +40,11 @@ resource "aws_alb_listener" "https" {
     type             = "forward"
   }
 }
+
+# Using SNI to attach multiple certificates to the same load balancer
+resource "aws_alb_listener_certificate" "https" {
```
samkahchiin, line 45: How about the default certificates? 
Will it be automatically set to default with the presence of additional certs?

samkahchiin (approved):
LGTM 🎉

PR #154, opened 2023-02-01
RDS: Use custom name everywhere if passed

swiknaba:
When adding a 2nd DB to the same environment/project, this currently would fail:

> Error: creating Security Group (dbl-production-db): InvalidGroup.Duplicate: The security group 'dbl-production-db' already exists for VPC 'vpc-09ba3920de77c22f8'
>
> Error: creating DB Parameter Group: DBParameterGroupAlreadyExists: Parameter group dbl-production-postgres13 already exists
>
> Error: creating DB Parameter Group: DBParameterGroupAlreadyExists: Parameter group dbl-production-postgres14 already exists
>
> Error: creating RDS DB Subnet Group (dbl-production): DBSubnetGroupAlreadyExists: The DB subnet group 'dbl-production' already exists.

In the test that resulted in above errors, a `name` was explicitly passed (`"${local.project}-${local.environment}-datascience"`)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) 
 This project does not follow SemVer, since modules are independent of each other; thus, SemVer does not make sense. Changes are grouped per module.
 
 ## [unreleased]
+### RDS
+- consistenly use `local.name` instead of `project-environment`. By default, both are the same, but `local.name` can be overridden to allow for more flexibility; e.g. when one needs to launch multiple RDS instances for the same project/environment.
 
 ## [v2022.12.12]
 ## Autoscaling/ECS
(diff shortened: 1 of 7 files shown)


samkahchiin (approved):
LGTM. All the name are replaced in rds module. :)

samkahchiin (changes requested):
For `rds/iam.tf`:
> line 30

Also replace the project and environment name to `local.name` ? 
> line 60 description 

change it to `local.name` or maybe we don't need it if it is replica?

swiknaba:
Thanks for the feedback, I've improved all descriptions and name tags now!


> * Do we actually need to create policy for accessing replica DB? Will we access it? 🤔

this PR is not about read-replicas, so I won't touch anything related to this. But likely, we never need to (manually) access read-replicas. We can remove that in a new PR.

This PR aims to add a separated second DB to an existing app, e..g. if we want to run write-heavy async imports that else would degrade the application's database's performance too much.

@samkahchiin PTAL 🙇

samkahchiin (approved):
LGTM. :)

Issue #153, opened 2023-01-09
Use Regex to determine the RDS version


swiknaba:
https://developer.hashicorp.com/terraform/language/functions/regex

https://github.com/dbl-works/terraform/blob/main/rds/rds.tf#L14

If we were to specify a minor version, e.g., `14.5` then `var.engine_version == "14"` would evaluate to "false", and use the v13 parameter group, which then results in an error.

To the user, it is not obvious, that a minor version cannot be specified, especially since passing a minor version is generally supported by RDS.

PR #155, opened 2023-02-03
[Chore] Support patch version in RDS module

samkahchiin:
### Descriptions
https://github.com/dbl-works/terraform/blob/main/rds/rds.tf#L14

If we were to specify a minor version, e.g., 14.5 then var.engine_version == "14" would evaluate to "false", and use the v13 parameter group, which then results in an error.

To the user, it is not obvious, that a minor version cannot be specified, especially since passing a minor version is generally supported by RDS.

### Relevant issues
resolves https://github.com/dbl-works/terraform/issues/153

### DIFF
--- a/rds/rds.tf
+++ b/rds/rds.tf
@@ -1,9 +1,13 @@
+locals {
+  major_engine_version = split(".", var.engine_version)[0]
+}
+
 resource "aws_db_instance" "main" {
   db_subnet_group_name                = aws_db_subnet_group.main.name
   allocated_storage                   = var.allocated_storage
   storage_type                        = "gp2"
   engine                              = var.is_read_replica ? null : "postgres"
-  engine_version                      = var.is_read_replica ? null : var.engine_version
+  engine_version                      = var.is_read_replica ? null : local.major_engine_version # Use major version only to allow AWS to update the minor/patch version automatically
   instance_class                      = var.instance_class
   identifier                          = "${local.name}${var.is_read_replica ? "-read-replica" : ""}"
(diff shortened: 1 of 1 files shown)


swiknaba (approved):
Thanks! Stumbled upon this issue as well before.

Should we generally strip the minor version? This would allow us to not ignore the engine_version (which we do due to AWS auto-update of minor versions).

Since we are ignoring the engine_version, we currently cannot upgrade a major version via terraform. Major upgrades come with downtime though: https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_UpgradeDBInstance.Upgrading.html so maybe not a bad thing we can't "easily" perform these via terraform. WDYT?

image

samkahchiin:
> Should we generally strip the minor version? This would allow us to not ignore the engine_version (which we do due to AWS auto-update of minor versions).

Didn't know that the AWS will do the auto minor version update if we don't specify the minor version. Can't think of any use cases which a minor version update will cause harm to the apps. So I think it is a good idea to strip this.

> Since we are ignoring the engine_version, we currently cannot upgrade a major version via terraform. Major upgrades come with downtime though

I think the reason why the engine_version change is ignored is because we don't want the minor version update keep showing changes whenever we terraform apply. 
Since we agree with the first point (stripping the version to the first octet), I assume this problem can be avoided and we can take out the engine_version from the ignore_changes ? 
Personally, I would prefer to be able to do the major version update thru terraform as user would not expect to go thru the hassle after using this module to create RDS. 🤔 WDYT?

swiknaba (approved):
I'm fine with removing the engine version ignore either now or later.

PR #158, opened 2023-02-06
[Chore] Make S3 replicas as optional variables in stack/app

samkahchiin:
### Description
Make S3 replicas as optional variables in stack/app

### DIFF
--- a/stack/app/README.md
+++ b/stack/app/README.md
@@ -82,7 +82,7 @@ module "stack" {
       bucket_name                     = "project-staging-media",
       versioning                      = true,
       primary_storage_class_retention = 0,
-      replicas                        = []
+      # replicas is an optional variables, it can be omitted
     },
   ]
 
--- a/stack/app/README.md
+++ b/stack/app/README.md
@@ -82,7 +82,9 @@ module "stack" {
       bucket_name                     = "project-staging-media",
       versioning                      = true,
       primary_storage_class_retention = 0,
-      # replicas is an optional variables, it can be omitted
+
+      # optional      
+      replicas = []
     },
   ]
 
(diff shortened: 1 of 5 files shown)


stack/app/README.md
```diff
@@ -82,7 +82,7 @@ module "stack" {
       bucket_name                     = "project-staging-media",
       versioning                      = true,
       primary_storage_class_retention = 0,
-      replicas                        = []
+      # replicas is an optional variables, it can be omitted
```
swiknaba, line 85:
```suggestion

      # optional      
      replicas = []
```

swiknaba: to be consistent with other readmes. Also easier to read and shows the type (array).

stack/app/variables.tf
```diff
@@ -59,11 +59,11 @@ variable "private_buckets_list" {
     bucket_name                     = string
     versioning                      = bool
     primary_storage_class_retention = number
-    replicas = list(object({
+    replicas = optional(list(object({
       bucket_arn = string
       kms_arn    = string
       region     = string
-    }))
+    })))
```
swiknaba, line 66:
```suggestion
    })), [])
```

does this work to have a default, so we do not need to compare to null everywhere? https://developer.hashicorp.com/terraform/language/expressions/type-constraints#optional-object-type-attributes

stack/app/variables.tf
```diff
@@ -76,9 +76,9 @@ variable "public_buckets_list" {
     bucket_name                     = string
     versioning                      = bool
     primary_storage_class_retention = number
-    replicas = list(object({
+    replicas = optional(list(object({
       bucket_arn = string
-    }))
+    })))
```
swiknaba, line 81: same here, if that works

swiknaba (approved):
Thanks for updating the changelog! Left non-blocking nit picks (not sure, if a default for a non-primitive object works)

stack/app/variables.tf
samkahchiin: Works 🎉 
Thanks for the recommendations!

swiknaba (approved):
👍

PR #169, opened 2023-02-15
[Chore] ecr: Add lifecycle policy

samkahchiin:
#### Summary
Add lifecycle policy so ECR images will be expired after some requirements are fulfilled

#### Motivation
Cut down AWS costs


swiknaba (approved):
image

PR #177, opened 2023-03-17
[Chore] Script for creating read_only DB role 

samkahchiin:

### DIFF
--- a/script/database_role/README.md
+++ b/script/database_role/README.md
@@ -1,0 +1,20 @@
+# Terraform Module: DB SQL Script - Create readonly role
+
+Run SQL script in DB to create DB role(s).
+This script will be run on the machine which run this TF script.
+
+## Usage
+
+```terraform
+module "db_readonly_script" {
+  source = "github.com/dbl-works/terraform//script/database_role?ref=v2023.03.06"
+
+  project = local.project
+  environment = local.environment
+  domain_name = "example.com"
+  db_identifier = "unique_db_name"
+
+  # Optional
+  bastion_subdomain = "bastion-staging"
+}
+```
--- a/script/database_role/db_readonly_role.sql
+++ b/script/database_role/db_readonly_role.sql
@@ -1,0 +1,7 @@
+CREATE ROLE {project}_{environment}_readonly;
+GRANT CONNECT ON DATABASE {db_name} TO {project}_{environment}_readonly;
+GRANT USAGE ON SCHEMA public TO {project}_{environment}_readonly;
+GRANT SELECT ON ALL TABLES IN SCHEMA public TO {project}_{environment}_readonly;
+ALTER DEFAULT PRIVILEGES IN SCHEMA public GRANT SELECT ON TABLES TO {project}_{environment}_readonly;
+GRANT rds_iam TO {project}_{environment}_readonly;
+ALTER ROLE {project}_{environment}_readonly WITH LOGIN;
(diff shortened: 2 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -36,6 +36,7 @@ jobs:
           - s3
           - s3-private
           - s3-public
+          - script/database_role
```
swiknaba, line 39: NIT: should we name it `database-roles`
* plural, since we can create scripts for different roles (we usually have a read-only role and and application-role)
* dashes for consistency (see `vpc-peering`, `s3-*`, `iam-for-*`)

script/database_role/README.md
```diff
@@ -0,0 +1,20 @@
+# Terraform Module: DB SQL Script - Create readonly role
+
+Run SQL script in DB to create DB role(s).
+This script will be run on the machine which run this TF script.
+
+## Usage
+
+```terraform
+module "db_readonly_script" {
```
swiknaba, line 9: Could we use this module in the RDS module? E.g. pass as a variable sth like `read_only_role: Boolean` (default `true`?) to create the read-only role when we create the DB instance?

this will reduce the amount of extra steps for each new stack.

script/database_role/db_readonly_role.sql
```diff
@@ -0,0 +1,7 @@
+CREATE ROLE {project}_{environment}_readonly;
```
swiknaba, line 1: great thanks! May I request a 2nd role: `db-app` which grants everything a web-app would need: read/write/delete/... basically any regular operation except for the root-level dangerous stuff such as `drop db`, `truncate`, ...

script/database_role/README.md
```diff
@@ -0,0 +1,20 @@
+# Terraform Module: DB SQL Script - Create readonly role
+
+Run SQL script in DB to create DB role(s).
+This script will be run on the machine which run this TF script.
```
swiknaba, line 4: NIT: shall we add that you need a SSH key locally that is allowlisted on the Bastion Host?

swiknaba (commented):
Generally, LGTM! Left a few suggestions, let me know what you think :)

script/database_role/README.md
samkahchiin: I think this would not reduce the number of extra steps because we still need to run the script afterwards. eg.
1. stack/app (for creation of ecs <bastion> and rds)
2.  deploy bastion
3. Run db script

script/database_role/db_readonly_role.sql
samkahchiin: @swiknaba  
How about this?

```
CREATE ROLE {project}_{environment}_app;
ALTER ROLE {project}_{environment}_app WITH LOGIN ENCRYPTED PASSWORD '{generate secure password}';

GRANT CONNECT ON DATABASE {project}_{environment} TO {project}_{environment}_app;
GRANT USAGE ON SCHEMA public TO {project}_{environment}_app;
GRANT ALL ON SCHEMA public TO {project}_{environment}_app;
GRANT ALL ON ALL SEQUENCES IN SCHEMA public TO {project}_{environment}_app;
GRANT INSERT, SELECT, UPDATE, DELETE, REFERENCES, TRIGGER ON ALL TABLES IN SCHEMA public TO {project}_{environment}_app;
```

script/database_role/README.md
swiknaba: Could the RDS module output a reminder to run this?

swiknaba (approved):

PR #186, opened 2023-03-24
rds: Add skip_final_snapshot variables for rds module

thekcsam:
#### Summary
In case of accidental deletion of rds, we should at least keep the last automated backup. 
In case we would like to keep the last copy of the rds forever, we should manually copy the snapshot (manual snapshot does not expire)
- https://aws.amazon.com/about-aws/whats-new/2018/11/amazon-rds-automated-backups-can-now-be-retained/
- https://repost.aws/knowledge-center/rds-automated-snapshot-retain-longer

#### Motivation
Safety measures for RDS

### DIFF
--- a/rds/rds.tf
+++ b/rds/rds.tf
@@ -1,6 +1,6 @@
 locals {
   major_engine_version      = split(".", var.engine_version)[0]
-  final_snapshot_identifier = "rds-${var.project}-${var.environment}-${formatdate("DD-MM-YY-hhmm", timestamp())}"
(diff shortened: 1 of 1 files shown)


samkahchiin:
@swiknaba 
The previous configuration will skip the final snapshot by default, 
but I think by default it should be set to false,
so we clearly understand that we don't need the snapshot when we set it explicitly set it (eg. staging DB)

swiknaba (approved):

PR #188, opened 2023-03-29
ecs_service: Add new module

samkahchiin:

### DIFF
--- a/ecs_service/ecs_task_definition.tf
+++ b/ecs_service/ecs_task_definition.tf
@@ -22,15 +22,15 @@ locals {
     CONTAINER_PORT        = var.app_container_port
     CONTAINER_NAME        = var.container_name
     ECR_REPO_NAME         = var.ecr_repo_name
-    ENVIRONMENT           = local.environment
+    ENVIRONMENT           = var.environment
     ENVIRONMENT_VARIABLES = jsonencode(local.environment_variables)
     IMAGE_TAG             = var.image_tag
     LOG_PATH              = var.log_path
     LOGGER_CONTAINER_PORT = var.logger_container_port
     LOGGER_ECR_REPO_NAME  = var.logger_ecr_repo_name
-    PROJECT               = local.project
+    PROJECT               = var.project
     REGION                = data.aws_region.current.name
-    VOLUME_NAME           = local.volume_name
+    VOLUME_NAME           = var.volume_name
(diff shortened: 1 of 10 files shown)


ecs_service/ecs_task_definition.tf
```diff
@@ -0,0 +1,58 @@
+data "aws_secretsmanager_secret" "app" {
+  name = "${var.project}/app/${var.environment}"
```
swiknaba, line 2:
```suggestion
  name = var.secrets_alias === null ? "${var.project}/app/${var.environment}" : var.secrets_alias
```

could we do sth like that to allow passing a alias that differs from the standard?

ecs_service/variables.tf
```diff
@@ -0,0 +1,74 @@
...
+}
+
+variable "image_tag" {
+  type    = string
+  default = "latest-main"
+}
+
+variable "cpu" {
+  type    = number
+  default = 512
```
swiknaba, line 16: should we default to the smallest possible, i.e. `256`?

ecs_service/variables.tf
```diff
@@ -0,0 +1,74 @@
...
+}
+
+variable "memory" {
+  type    = number
+  default = 1024
```
swiknaba, line 21: should we default to the smallest possible, i.e. `512`?

ecs_service/variables.tf
```diff
@@ -0,0 +1,74 @@
...
+  description = "path in the apps which store the log"
+}
+
+# TODO: Read env in a file
+variable "environment_variables" {
+  type    = map(string)
+  default = {}
+}
+
+variable "secrets" {
```
swiknaba, line 70: NIT: name it `secrets_names`?

iam/iam-for-deploy-bot/main.tf
```diff
@@ -352,15 +114,47 @@ resource "aws_iam_group" "deploy-bot-deploy-access" {
   name = "deploy-bot-deploy-access"
 }
 
+resource "aws_iam_policy" "deploy-bot-extra-access" {
+  name        = "DeployBot_ExtraAccess"
+  path        = "/"
+  description = "Allow extra access for deploy-bot"
```
swiknaba, line 120: NIT: can we document here **why** we need this additional access? Maybe we could also find a better name than "extra" to clarify what this is for.

swiknaba (approved):
sloth

Issue #171, opened 2023-02-17
Send slack message when a vulnerability is found in the ECR image scanning


samkahchiin:
References: 
- https://dev.to/hayao_k/amazon-ecr-image-scan-results-with-slack-notification-2ab1
- https://docs.aws.amazon.com/AmazonECR/latest/userguide/ecr-eventbridge.html

PR #196, opened 2023-04-11
slack/ecr-scanner-notifier: Add ECR scanner Slack notifier

samkahchiin:
#### Summary
Send slack message when a vulnerability is found in the ECR image scanning 

#### Motivation
closes https://github.com/dbl-works/terraform/issues/171


swiknaba (commented):
I like this idea. Using a Lambda seems to be a good choice. Any chance you'll have time to finish this? Low prior I think. Also: why add python (yet another tool) to our stack?

slack/ecr-scanner-notifier/main.tf
```diff
@@ -12,29 +15,70 @@ module "lambda" {
...
+        "HIGH": [{
+          "exists": true
+        }]
+      }, {
+        "MEDIUM": [{
+          "exists": true
+        }]
+      }, {
+        "LOW": [{
+          "exists": true
```
swiknaba, line 48: Should we ignore "low"?

slack/ecr-scanner-notifier/script/main.rb
```diff
@@ -0,0 +1,82 @@
...
+
+def build_slack_message(event)
+  account_id = event['account']
+  detail = event['detail']
+  region = ENV['AWS_DEFAULT_REGION']
+  channel = ENV['CHANNEL']
+
+  message = "*ECR Image Scan findings | #{region} | Account ID:#{account_id}*"
+  repository_name = detail['repository-name']
+  severity_list = detail['finding-severity-counts']
```
swiknaba, line 23:
```suggestion
  severity_counts = detail['finding-severity-counts']
```

NIT

slack/ecr-scanner-notifier/script/main.rb
```diff
@@ -0,0 +1,82 @@
...
+  text_properties = get_properties(severity_list)
```
swiknaba, line 24:
```suggestion
  text_properties = text_block_from(severity_counts)
```

NIT

slack/ecr-scanner-notifier/script/main.rb
```diff
@@ -0,0 +1,82 @@
+require 'json'
+require 'net/http'
+require 'uri'
+
+def get_properties(finding_counts)
```
swiknaba, line 5:
```suggestion
def text_block_from(finding_counts)
```

NIT

slack/ecr-scanner-notifier/script/main.rb
```diff
@@ -0,0 +1,82 @@
...
+    'icon_emoji' => ':ecr:',
+    'text' => message,
+    'attachments' => [
+      {
+        'fallback' => 'AmazonECR Image Scan Findings Description.',
+        'color' => text_properties['color'],
+        'title' => "#{text_properties['icon']} #{repository_name}:#{detail['image-tags'][0]}",
+        'title_link' => "https://#{region}.console.aws.amazon.com/ecr/repositories/#{repository_name}/_/image/#{detail['image-digest']}/scan-results?region=#{region}",
+        'text' => "Image Scan Completed at #{event['time']}",
+        'fields' => severity_list.map { |k, v| { 'title' => k.capitalize, 'value' => v, 'short' => true } }
```
swiknaba, line 38:
```suggestion
        'fields' => severity_counts.map { |severity_level, count| { 'title' => severity_level.capitalize, 'value' => count, 'short' => true } }
```

NIT

lambda/README.md
```diff
@@ -9,14 +9,16 @@ Used for managing lambda functions.
 module "lambda" {
   source = "github.com/dbl-works/terraform//lambda?ref=main"
 
-  project     = "dbl"
-  environment = "production"
-  source_dir  = "Path to the directory containing the lambda function code."
+  function_name   = "lambda"
+  project         = "dbl"
+  environment     = "production"
+  source_dir = "Path to the directory containing the lambda function code."
```
swiknaba, line 15:
```suggestion
  environment     = "production"
  source_dir      = "Path to the directory containing the lambda function code."
```

NIT

lambda/README.md
```diff
@@ -9,14 +9,16 @@ Used for managing lambda functions.
...
 
   # optional
   handler     = "index.handler"
   timeout     = 10
   memory_size = 1024
+  runtime     = "ruby3.2"
```
swiknaba, line 21: ⚠️ default value in the variables files is `nodejs16.x`. We should have the same value in the README and the defaults.

slack/ecr-scanner-notifier/script/main.rb
```diff
@@ -0,0 +1,82 @@
...
+  #          "CRITICAL": 10,
+  #          "MEDIUM": 9
+  #        },
+  #       "image-digest": "sha256:7f5b2640fe6fb4f46592dfd3410c4a79dac4f89e4782432e0378abcd1234",
+  #       "image-tags": ["commit-xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"]
+  #   }
+  # }
+  #
+  slack_message = build_slack_message(event)
+  uri = URI(ENV['WEBHOOK_URL'])
```
swiknaba, line 70:
```suggestion
  uri = URI(ENV.fetch('WEBHOOK_URL'))
```

If the ENV var is not set and you pass `nil` to `URI` you get the error 

> `URI': bad argument 

which is harder to debug than

> key not found: "WEBHOOK_URL"

which instantly tells you what is the underlying problem


💡 general rule: **always** use `.fetch` ‼️. You _may_ use `[]` as a shortcut for `.fetch('key', nil)` when you intend that `nil` is a proper fallback value, and your code is built to proceed with this fallback value.

slack/ecr-scanner-notifier/script/main.rb
```diff
@@ -0,0 +1,82 @@
...
+  req = Net::HTTP::Post.new(uri, 'Content-Type' => 'application/json')
+  req.body = slack_message.to_json
+
+  begin
+    Net::HTTP.start(uri.hostname, uri.port, use_ssl: uri.scheme == 'https') do |http|
```
swiknaba, line 75: Can we enforce using SSL for security? If someone pastes a non-SSL URL, we should treat it as a defect.

slack/ecr-scanner-notifier/script/main.rb
```diff
@@ -0,0 +1,82 @@
...
+      http.request(req)
+    end
+    puts('Message posted.')
+  rescue StandardError => e
+    puts("Request failed: #{e.message}")
```
swiknaba, line 80: Where do we log these? Do we have a log-group? Does the module output the log-group-name, so I can document it?

slack/ecr-scanner-notifier/variables.tf
```diff
@@ -0,0 +1,11 @@
+variable "project" {
+  type = string
+}
+
+variable "slack_webhook_url" {
+  type = string
+}
```
swiknaba, line 7: Can we add a validation that fails, if the URL does not start with `https` to enforce SSL?

stack/global/variables.tf
```diff
@@ -132,3 +132,11 @@ variable "private_buckets_list" {
   }))
   description = "S3 private buckets that are required globally, such as buckets used for storing Terraform state files or centralized log storage."
 }
+
+variable "ecr_scanner_notifier_config" {
```
swiknaba, line 136: Nice!

swiknaba (approved):
Nothing blocking, but if you have, time please ensure:
* the documentation in the README is not different from the default value in the variables file
* never use `[]`, unless a `nil` return value is a tested code path

I've also left a few NITs that help make the code more readable.

slack/ecr-scanner-notifier/script/main.rb
samkahchiin: Yes. it is the default lambda log group name. will add it in the output. :)

swiknaba (approved):
Awesome feature, thank you!

swiknaba (approved):
fixed the name of the lambda output :)

PR #198, opened 2023-04-13
aws-transfer: new module

samkahchiin:
#### Summary
This module creates the following:
- AWS transfer family service
- AWS transfer family users
- Private S3 bucket (encrypted by KMS key)

It should not 
- [x] allow the user to delete the file
- [x] allow the user to cd to the root folder


swiknaba (approved):
Tested on our test AWS account. All works 👍 ✅

PR #199, opened 2023-04-16
Allow logger side-cear container to access secrets

swiknaba:
#### Summary

Allow granting access to secrets to the logger container

#### Motivation

I want to move API keys for external log sinks out of the vector config file (git tracked) to AWS Secrets Manager.

Example from a vector config file:

```toml
  [sinks.logtail_http_sink.auth]
    strategy = "bearer"
    token    = "${LOGTAIL_VECTOR_SINK_TOKEN}"
```

For this to work, we need to grant the logger container access to some (but not all) secrets.

We should use a different `secrets` variable for this, since the logger should not have access to all secrets; it is not necessary and can only lead to bugs where we inject sensitive data into the logs by accident.

### DIFF
--- a/ecs_service/README.md
+++ b/ecs_service/README.md
@@ -24,6 +24,7 @@ module "ecs_service" {
   logger_image_tag       = "latest-main"
   memory                 = 512
   secrets                = []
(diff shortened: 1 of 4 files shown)


samkahchiin (approved):
👍

PR #203, opened 2023-04-21
Allow force-redeployment of services that have no changes

swiknaba:
#### Summary

Introduces an updateable tag so we can force redeploy e.g. bastion on each deploy (when the bastion image tag hasn't changed)


swiknaba:
we can simple inject `var.app_image_tag` as a ENV VAR into the tasks that we want to force-redeploy

PR #206, opened 2023-05-01
Logging at scale

swiknaba:
* actually, log volume should be optional --> choose it for MVPs due to its simplicity (but don't use it at scale)
* kinesis as alternative to the log volume attached to the web task for apps at scale

also see https://github.com/dbl-works/terraform/pull/205

closes https://github.com/dbl-works/terraform/pull/205

### DIFF
--- a/cloudwatch-kinesis/firehose-delivery-stream.tf
+++ b/cloudwatch-kinesis/firehose-delivery-stream.tf
@@ -46,7 +46,8 @@ resource "aws_iam_policy" "kinesis" {
         "Action" : [
           "logs:DescribeLogGroups",
           "logs:DescribeLogStreams",
-          "logs:GetLogEvents"
+          "logs:GetLogEvents",
+          "logs:PutLogEvents"
         ],
         "Resource" : "*"
       },
--- a/cloudwatch-kinesis/firehose-delivery-stream.tf
+++ b/cloudwatch-kinesis/firehose-delivery-stream.tf
@@ -15,26 +15,26 @@ resource "aws_kinesis_firehose_delivery_stream" "main" {
   destination = "http_endpoint"
 
   s3_configuration {
-    role_arn           = aws_iam_role.ecs-task-execution.arn
-    bucket_arn         = variable.log_bucket_arn
-    buffer_size        = 10
-    buffer_interval    = 400
+    role_arn           = data.aws_iam_role.ecs-task-execution.arn
+    bucket_arn         = var.log_bucket_arn
+    buffer_size        = var.buffer_size_for_s3     # in MB
+    buffer_interval    = var.buffer_interval_for_s3 # in seconds
     compression_format = "GZIP"
   }
 
   http_endpoint_configuration {
     url                = "${data.aws_lb.main.dns_name}:${var.ecs_http_port}"
     name               = local.ecs_cluster_name
-    buffering_size     = 1   # MB
-    buffering_interval = 600 # 1 minute
-    role_arn           = aws_iam_role.ecs-task-execution.arn
+    buffering_size     = var.buffer_size_for_http_endpoint
+    buffering_interval = var.buffer_interval_for_http_endpoint
+    role_arn           = data.aws_iam_role.ecs-task-execution.arn
     s3_backup_mode     = "FailedDataOnly"
 
     request_configuration {
       content_encoding = "GZIP"
 
       common_attributes {
-        name  = "envionment"
+        name  = "environment"
         value = var.environment
       }
 
--- a/cloudwatch-kinesis/firehose-delivery-stream.tf
+++ b/cloudwatch-kinesis/firehose-delivery-stream.tf
@@ -58,8 +58,9 @@ resource "aws_iam_policy" "kinesis" {
           "kms:Decrypt",
           "kms:GenerateDataKey"
         ],
-        # TODO: Please stricten the rules
-        "Resource" : "*"
+        "Resource" : [
+          var.s3_kms_arn
+        ]
       }
     ]
   })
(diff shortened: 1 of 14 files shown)


cloudwatch-kinesis/variables.tf
```diff
@@ -0,0 +1,32 @@
...
+variable "region" {
+  type    = string
+  default = "eu-central-1"
+}
+
+variable "ecs_cluster_name" {
+  type = string
+}
+
+variable "region" {
```
samkahchiin, line 24: duplicated region

cloudwatch-kinesis/variables.tf
```diff
@@ -0,0 +1,32 @@
...
+}
+
+
+# Regional allows clusters with the same name to be in multiple regions
+variable "region" {
+  type    = string
+  default = "eu-central-1"
+}
+
+variable "ecs_cluster_name" {
```
samkahchiin, line 20: should set default to null to make it optinal

cloudwatch-kinesis/variables.tf
```diff
@@ -0,0 +1,67 @@
...
+
+variable "buffer_size_for_s3" {
+  type        = number
+  default     = 10
+  description = "Buffer incoming data to the specified size, in MBs, before delivering it to the s3 bucket."
+}
+
+variable "buffer_interval_for_s3" {
+  type        = number
+  default     = 400
```
swiknaba, line 43:
```suggestion
  default     = 1800 # 30 min
```

1 file every 6-7 minutes is already a lot, especially if we run a lot of tasks. Imagine we run 10 tasks, each logging 1 file every 6.5 minutes:

10 tasks * 1 file / 6.5 minutes * 1440 minutes/day * 30.25 days/month = 67.000 files per month.

Can we prefix the path for the files in the S3 bucket? With vector, we had the following prefix:

```
"${RAILS_ENV}/%Y-%m/%d/%H/%F-%H:%M-${RAILS_ENV}-"
```

resulting in:
* 1 folder per environment, 1 folder per year/month, 1 folder per day, 1 folder per hour
* files are prefixed with the timestamp and the environment

that should make it fairly easy to retrieve log files manually for a given time period. WDYT?

ecs-deploy/locals.tf
```diff
@@ -31,12 +31,6 @@ locals {
   region            = data.aws_region.current.name
   image_name        = var.app_image_name == null ? "${local.account_id}.dkr.ecr.${local.region}.amazonaws.com/${var.ecr_repo_name}" : var.app_image_name
   logger_image_name = var.with_logger && var.logger_image_name == null ? "${local.account_id}.dkr.ecr.${local.region}.amazonaws.com/${var.logger_ecr_repo_name}" : var.logger_image_name
-  mount_points = var.with_logger ? [
```
swiknaba, line 34: Should we optionally keep this for projects that have very low request volume, so we don't need to set up Kinesis and all that comes with it? Collecting the logs with vector on the same container is much simpler (it just doesn't scale, which is not important for every project) --> I'm also fine with removing this and setting up Kinesis for all projects 🤷  --- WDYT @samkahchiin

cloudwatch-kinesis/firehose-delivery-stream.tf
```diff
@@ -0,0 +1,153 @@
...
+    ]
+  })
+}
+
+resource "aws_iam_role_policy_attachment" "kinesis" {
+  role       = aws_iam_role.kinesis.name
+  policy_arn = aws_iam_policy.kinesis.arn
+}
+
+resource "aws_kinesis_firehose_delivery_stream" "main" {
```
swiknaba, line 73: Should we add a 2nd (optional) stream to also send all logs to a S3 bucket for long term storage? @samkahchiin (using the `extended_s3` endpoint)

swiknaba (commented):
Generally, this works. Left a few comments on where I think we can improve this.

cloudwatch-kinesis/variables.tf
samkahchiin: We might not have a file generated for each task because the file is only generated when there is an error in http-endpoint/processing failed

Screenshot 2023-05-04 at 12 19 55 PM

In fact, I think we should have a shorter time to get an alert as soon as possible when the error happens. (Perhaps to send Slack message when a new file is added to the bucket?)

> Can we prefix the path for the files in the S3 bucket? 

Yes!

Screenshot 2023-05-04 at 12 23 02 PM

samkahchiin: @swiknaba 
Should we also backup the logs in S3?

ecs-deploy/locals.tf
samkahchiin: Agree. I think we should keep it before we find another easier solution. 
Maintaining kinesis for all projects can be tiring.

cloudwatch-kinesis/firehose-delivery-stream.tf
samkahchiin: ahhh. This comment is hidden in the front page.
I am thinking about this as well. 
It can be achieved by setting the mode to AllData, so we don't need the extended_s3 endpoint

cloudwatch-kinesis/variables.tf
```diff
@@ -0,0 +1,93 @@
+locals {
+  ecs_cluster_name = var.ecs_cluster_name == null ? "${var.project}-${var.environment}" : var.ecs_cluster_name
+  name             = "${var.project}-${var.environment}-http-endpoint"
+  log_group_name   = "kinesis/${local.ecs_cluster_name}"
```
samkahchiin, line 4: @swiknaba should this be /kinesis ? 
I saw we have a pattern to prepend a / in front of the log group but i don't know why. 😆

s3-private/s3.tf
```diff
@@ -40,6 +40,16 @@ resource "aws_s3_bucket_acl" "main-bucket-data-acl" {
   acl    = "private"
 }
 
+# https://aws.amazon.com/about-aws/whats-new/2021/11/amazon-s3-object-ownership-simplify-access-management-data-s3/
+# https://github.com/hashicorp/terraform-provider-aws/issues/22271
+resource "aws_s3_bucket_ownership_controls" "main" {
```
samkahchiin, line 45: @swiknaba  
I keep getting an error when creating a s3 private bucket (which hasn't happened before). 
This solution is suggested by the issues thread but doesn't solve the problem completely (I have to update the permission in the  console to completely get rid of the error)

cloudwatch-kinesis/firehose-delivery-stream.tf
swiknaba: That sounds good to me

swiknaba (commented):
I've done some more testing on our test app with this, and logging generally works quite well. I vote we merge this PR, it's already huge, and before we release the next version, we clean this up if necessary. WDYT @samkahchiin

ecs-deploy/locals.tf
samkahchiin: @swiknaba 

Putting back mountpoint in this commit => 23e13d5

samkahchiin:
Yes. I would also prefer to merge this before it goes even larger. 👍

PR #213, opened 2023-05-14
Ensure the correct parameter group is used, regardless of engine version

swiknaba:
#### Summary

Previously, if the engine version was not 14, version 13 was assumed for the parameter group.

This obviously does not allow creating new instances using any other version.

#### Motivation

Postgres 15 was released on AWS.

### DIFF
--- a/rds/rds.tf
+++ b/rds/rds.tf
@@ -59,6 +59,8 @@ resource "aws_db_instance" "main" {
       engine_version, # AWS will auto-update minor version changes
       db_name,        # if you didn't use this before it would re-create your RDS instance
       snapshot_identifier,
+      username,
+      password,
     ]
   }
 }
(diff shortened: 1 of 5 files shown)


rds/rds.tf
```diff
@@ -1,9 +1,12 @@
 locals {
   major_engine_version      = split(".", var.engine_version)[0]
   final_snapshot_identifier = "final-snapshot-${var.project}-${var.environment}-${formatdate("DD-MM-YY-hhmm", timestamp())}"
+  # @NOTE: `"default.postgres${local.major_engine_version}"` does not exist for e.g. postgres15 eventhough, postges 15 is generally available
```
samkahchiin, line 4: Can we skip the creations of the 13/14 parameter group if the engine_version is not 13/14?

samkahchiin (changes requested):

samkahchiin (changes requested):
Please also update CHANGELOG

samkahchiin (approved):
Test by running terraform apply in an existing rds module => works. 👍

PR #214, opened 2023-05-15
ecs-deploy: rename variables

samkahchiin:
#### Summary
1. ecs-deploy
- :warning: breaking change: add/remove/modify variable names
- allow user to add multiple sidecar task definitions

2. iam/iam-for-deploy-bot
- add permissions to manage logs

3. rds
- remove timestamp from final_snapshot_identifier so rds is not force updated in every deployment

4. fivetran/connectors
- fix bug in aws_iam_role_policy_attachment

5. fivetran/connectors/lambda
- Add sync_frequency as variable


ecs-deploy/locals.tf
```diff
@@ -6,76 +6,66 @@ data "aws_caller_identity" "current" {}
 data "aws_region" "current" {}
 
 locals {
+  name       = var.cluster_name != null ? var.cluster_name : "${var.project}-${var.environment}${var.regional ? "-${var.region}" : ""}"
+  account_id = data.aws_caller_identity.current.account_id
+  region     = data.aws_region.current.name
```
swiknaba, line 11: where is the region read from? Assume I have set a specific region the terraform/aws provider, but my local shell has a different `AWS_REGION` env var set would that result in a conflict?

swiknaba (approved):
Generally LGTM. Since we removed the fivetran AWS account ID from the readme: shall we simply hardcode it everywhere?

ecs-deploy/locals.tf
samkahchiin: it is read from the aws profile which run the terraform script
1. If you have aws provider and specify the region -> this region would be used
2. If you don't have aws provider but you have env var -> it will use the default profile

swiknaba (approved):

swiknaba (approved):

fivetran/README.md
```diff
@@ -70,7 +69,7 @@ module "fivetran" {
   lambda_settings = {
     lambda_role_arn = "arn:aws:iam::123456789:role/fivetran_lambda" # optional
     lambda_role_name = "fivetran_lambda" # optional
-    fivetran_aws_account_id ="834469178297" # optional
```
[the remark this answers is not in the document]
samkahchiin: fivetran_aws_account_id is not passed as a key in lambda_settings anymore.

It is still a variable in iam-policy-for-fivetran-lambda => https://github.com/dbl-works/terraform/tree/main/iam/iam-policy-for-fivetran-lambda

PR #215, opened 2023-05-15
cloudwatch-kinesis: add regional to allow deployment for regional service

samkahchiin:
#### Summary
- Add regional variable to allow deployment for regional service
- Add subscription_filter_pattern as variable


swiknaba (approved):
🎉

PR #218, opened 2023-05-18
s3-public: fix public bucket access ACL

samkahchiin:
#### rds
- Add identifier as variable

#### s3-public
- Add public bucket configuration to avoid ACCESS DENIED error. (https://github.com/hashicorp/terraform-provider-aws/issues/31021)


swiknaba (approved):
🚀 🧙‍♀️

samkahchiin:
NOTE for @swiknaba 
Following the suggestion in the issue still doesn't completely solve the problem.
We still need to set ACL in the console.

PR #228, opened 2023-07-11
ecs-deploy: allow the deployment of ecs service to private subnets

samkahchiin:
#### Summary
Allow the deployment of ecs service to private subnets

### DIFF
--- a/ecs-deploy/ecs_service.tf
+++ b/ecs-deploy/ecs_service.tf
@@ -42,6 +42,19 @@ data "aws_subnets" "public" {
   }
 }
 
+data "aws_subnets" "private" {
+  filter {
+    name   = "vpc-id"
+    values = [data.aws_vpc.main.id]
+  }
+  filter {
+    name = "tag:Name"
+    values = [
+      "${var.project}-${var.environment}-private-*",
+    ]
+  }
+}
+
 data "aws_security_group" "ecs" {
   name = "${local.name}-ecs"
 }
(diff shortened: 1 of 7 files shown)


ecs-deploy/ecs_service.tf
```diff
@@ -42,6 +42,19 @@ data "aws_subnets" "public" {
...
 
+data "aws_subnets" "private" {
+  filter {
+    name   = "vpc-id"
+    values = [data.aws_vpc.main.id]
+  }
+  filter {
+    name = "tag:Name"
+    values = [
+      "${var.project}-${var.environment}-private-*",
```
swiknaba, line 53: NIT: should we just have `data "aws_subnets" "main" { ... }` and use  `"${var.project}-${var.environment}-${var.subnet_type}-*"`?

swiknaba (dismissed):
Thanks! Left some non-blocking suggestion to reduce boilerplate.

ecs-deploy/ecs_service.tf
samkahchiin: Good suggestion. 👍 
I have updated the codes

stack/app/nat.tf
```diff
@@ -1,6 +1,7 @@
 module "nat" {
-  source = "../../nat"
+  count = length(var.public_ips) > 0 ? 1 : 0
```
swiknaba, line 2: this is a breaking change. If there wasn't a public IP before, we would still create route tables. Are we sure, no app depends on having those route tables?

nat/nat.tf
```diff
@@ -28,7 +28,7 @@ resource "aws_route" "main" {
 
 # Create route tables so elastic IPs are used for outgoing traffic
 resource "aws_route_table" "main" {
-  count  = length(var.subnet_public_ids)
+  count  = length(var.public_ips)
```
swiknaba, line 31: we later iterate over (private) subnets and get the route table for each (private) subnet. Does the number of public IPs necessarily have to be the same as subnets? if yes, this change is pointless, if no, the route table association will fail since the count is different.

swiknaba (changes requested):

swiknaba (approved):

PR #229, opened 2023-07-19
Set Load Balancer keep alive timeout to 60 seconds

swiknaba:
#### Summary

The current setting of 10 minutes is too long. The default is 60, which we should keep.

Rails /w Puma apps even default to 20 seconds. If you were to use a Ruby/Puma server with its default settings, having the LB keep connections open for more than 20 seconds 
will lead to bugs.

Also see https://github.com/puma/puma/issues/957

The update can be done in-place
image

### DIFF
--- a/ecs/README.md
+++ b/ecs/README.md
@@ -4,7 +4,7 @@ Create a compute cluster for hosting docker based apps.
 
 ## Notes
 
-The `idle_timeout` is set to 60 seconds. Make sure that your webserver's keep-alive timeout is set to 60 seconds (or more) to prevent the load balancer from keeping connections open to that the server has already closed.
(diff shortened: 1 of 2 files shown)


ecs/README.md
```diff
@@ -2,7 +2,11 @@
 
 Create a compute cluster for hosting docker based apps.
 
+## Notes
 
+The `idle_timeout` is set to 60 seconds. Make sure that your webserver's keep-alive timeout is set to 60 seconds (or more) to prevent the load balancer from keeping connections open to that the server has already closed.
```
samkahchiin, line 7:
```suggestion
The `idle_timeout` is set to 60 seconds. Ensure that your webserver's keep-alive timeout is set to 60 seconds (or more) to prevent the load balancer from keeping connections open to the already closed server.
```

samkahchiin (changes requested):

samkahchiin (approved):

PR #231, opened 2023-07-21
enable in transit encryption for elasticache

swiknaba:
#### Summary

* create a custom parameter group, if `parameter_group` variable is omitted instead of checking `maxmemory_policy`, since that is more intuitive
* enforce `in-transit-encryption` for better security (requirement of ISO 27001, see https://www.dataguard.co.uk/blog/iso-27001-annex-a.10-cryptography)

https://docs.aws.amazon.com/AmazonElastiCache/latest/red-ug/in-transit-encryption.html

--> ⚠️ Enabling in-transit encryption is a *breaking* change, i.e., it requires to re-create the Redis cluster. If you need zero downtime, you must create a new cluster, swap endpoints in the app, then destroy the old one. Data loss will occur, though.

see also: https://aquasecurity.github.io/tfsec/v0.61.3/checks/aws/elasticache/enable-in-transit-encryption/

#### Motivation

security (specifically ISO 27001)

--> Tested upgrading from Redis 6 without encryption to Redis 7 with encryption in transit: works ✅ (but requires the cluster to be recreated).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ This project does not follow SemVer, since modules are independent of each other
 ### ecs-deploy
 - Add deployment_circuit_breaker options [#230](https://github.com/dbl-works/terraform/pull/230)
 
+### elasticache
+- Enable `encryption-in-transit` by default [#231](https://github.com/dbl-works/terraform/pull/231)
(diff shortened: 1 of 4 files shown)


samkahchiin (approved):
LGTM :D

PR #232, opened 2023-07-25
Make Keep Alive Configurable

swiknaba:
##### Summary

- make keep alive timeout configurable
- NLB also respects configured timeout


samkahchiin (commented):
Should we also add keep_alive_timeout to stack/app module?

PR #236, opened 2023-08-03
lambda: add lambda_role_name

samkahchiin:
#### Summary
Add lambda role name to lambda module

#### Motivation
- We can skip the creation of role in the module

### DIFF
--- a/lambda/iam.tf
+++ b/lambda/iam.tf
@@ -3,7 +3,7 @@ locals {
 }
 
 data "aws_iam_role" "main" {
-  name = var.lambda_role_name
+  name = var.lambda_role_name == null ? aws_iam_role.main[0].name : var.lambda_role_name
 }
 
 data "aws_iam_policy_document" "combined" {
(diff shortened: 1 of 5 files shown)


lambda/iam.tf
```diff
@@ -2,6 +2,10 @@ locals {
   secret_and_kms_policy_json = length(var.secrets_and_kms_arns) > 0 ? data.aws_iam_policy_document.main.json : data.aws_iam_policy_document.dummy.json
 }
 
+data "aws_iam_role" "main" {
+  name = var.lambda_role_name
```
swiknaba, line 6: does this work, when `var.lambda_role_name`?

if no, we could do sth like:


```
name = var.lambda_role_name == null ? aws_iam_role.main.name : var.lambda_role_name
```

and then you can continue to use `data.aws_iam_role.main` through the rest of the code and don't need to duplicate the ternary operator all the time.
```

swiknaba (approved):
overall LGTM; left a non-blocking suggestion to simplify the code.

lambda/iam.tf
samkahchiin: Great suggestions! 👍

swiknaba (approved):

PR #237, opened 2023-08-03
New Module: HTTP Proxy to assume a static IP

swiknaba:
#### Summary
- New Module 'HTTP Proxy' to assume a static IP
- fix readme
- update changelog

#### Motivation

A cheap and simple way to assume a static IP that does not need to scale at the level of a NAT.


samkahchiin (approved):
LGTM.
Please also add the new module to the github test

swiknaba:
> 

Good point, thanks! PTAL :)

PR #238, opened 2023-08-03
Secure Proxy by limiting egress traffic & improve its documentation

swiknaba:
* improve security by limiting egress traffic to what is required instead of all of the internet
* add documentation on how to configure Tinyproxy

### DIFF
--- a/http-proxy/README.md
+++ b/http-proxy/README.md
@@ -24,5 +24,89 @@ module "http-proxy" {
   instance_type = "t3.micro"
   ssh_enabled   = false                   # switch this to "true" for the initial configuration, but keep it "false" for production
   ami_id        = "ami-0502e817a62226e03" # ubuntu 20.04 which has a free quota
+
+  # protocol & cidr_block are optional and default to TCP and all IPs
+  # protocol: one of "tcp", "udp", "icmp", or "-1" for all protocols
+  egress_rules = [
+    {
+      port        = 22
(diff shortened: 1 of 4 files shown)


http-proxy/README.md
```diff
@@ -24,5 +24,89 @@ module "http-proxy" {
...
+ssh -i ${PATH_TO_SSH_KEY} ${SSH_USER}@${SERVER_HOST}
+
+# you should now be logged into the proxy server
+sudo apt-get update && sudo apt-get install -y tinyproxy && sudo /etc/init.d/tinyproxy start
+
+sudo apt-get --only-upgrade install tinyproxy
+
+# purge current config file
+sudo cat /dev/null > /etc/tinyproxy/tinyproxy.conf # manually delete in nano if you get permission denied
+sudo nano /etc/tinyproxy/tinyproxy.conf
```
samkahchiin, line 104: Why don't we just `sudo cp tinyproxy.conf /etc/tinyproxy/tinyproxy.conf` ?

samkahchiin (approved):
LGTM 👍

http-proxy/README.md
swiknaba: You'd need to use SCP since the one config is on local but you replace the file on remote. Looking forward, maybe we can at some point define our own custom AMI using e.g. packer and include the file there.

Feel free to add alternative ways to the docs :)

PR #242, opened 2023-08-09
Add SQL script to grant access to certain tables for Fivetran

puckzxz:
#### Summary

Adds an SQL script to create a write-only Fivetran role to allow syncing data from Fivetran to an RDS database

#### Motivation

Project need

### DIFF
--- a/fivetran/rds-writeonly-role.sql
+++ b/fivetran/rds-writeonly-role.sql
@@ -1,0 +1,26 @@
+-- creates a write-only role for fivetran to connect to RDS for specific tables
+DO
+$$
+  DECLARE
+    table_name     varchar;
+    project        varchar   := '';
+    environment    varchar   := '';
+    allowed_tables varchar[] := array [
+      'table1',
+      'table2'
+      ];
+    password       varchar   := 'my-super-secret-password';
+  BEGIN
+    EXECUTE format('CREATE ROLE %s_%s_writeonly_fivetran;', project, environment);
+    EXECUTE format('ALTER ROLE %s_%s_writeonly_fivetran WITH LOGIN ENCRYPTED PASSWORD ''%s'';', project, environment,
+                   password);
+    EXECUTE format('GRANT USAGE ON SCHEMA public TO %s_%s_writeonly_fivetran;', project, environment);
+    FOREACH table_name IN ARRAY allowed_tables
+      LOOP
+        EXECUTE format('GRANT CREATE, UPDATE ON TABLE %s TO %s_%s_writeonly_fivetran', table_name, project,
+                       environment);
+      END LOOP;
+    EXECUTE format('ALTER DEFAULT PRIVILEGES IN SCHEMA public GRANT SELECT ON TABLES TO %s_%s_writeonly_fivetran;',
+                   project, environment);
+  END
+$$;


fivetran/rds-writeonly-role.sql
```diff
@@ -0,0 +1,27 @@
...
+  BEGIN
+    EXECUTE format('CREATE ROLE %s_%s_writeonly_fivetran;', project, environment);
+    EXECUTE format('ALTER ROLE %s_%s_writeonly_fivetran WITH LOGIN ENCRYPTED PASSWORD ''%s'';', project, environment,
+                   password);
+    EXECUTE format('GRANT USAGE ON SCHEMA public TO %s_%s_writeonly_fivetran;', project, environment);
+    FOREACH table_name IN ARRAY allowed_tables
+      LOOP
+        EXECUTE format('GRANT ALTER ON TABLE %s TO %s_%s_writeonly_fivetran', table_name, project, environment);
+        EXECUTE format('GRANT CREATE, UPDATE ON TABLE %s TO %s_%s_writeonly_fivetran', table_name, project,
+                       environment);
```
swiknaba, line 22:
```suggestion
        EXECUTE format('GRANT INSERT, SELECT, UPDATE, DELETE, ALTER ON TABLE %s TO %s_%s_writeonly_fivetran', table_name, project, environment);
```

we can summarize these. We'll also need delete & select so the sync can correctly determine which records have been inserted and delete records that need to be removed.

fivetran/rds-writeonly-role.sql
```diff
@@ -0,0 +1,27 @@
...
+      END LOOP;
+    EXECUTE format('ALTER DEFAULT PRIVILEGES IN SCHEMA public GRANT SELECT ON TABLES TO %s_%s_writeonly_fivetran;',
```
swiknaba, line 24: I'd remove this. If we add new tables, we should update the script.

fivetran/rds-writeonly-role.sql
```diff
@@ -0,0 +1,27 @@
...
+    table_name     varchar;
+    project        varchar   := '';
+    environment    varchar   := '';
+    allowed_tables varchar[] := array [
+      'table1',
+      'table2'
+      ];
+    password       varchar   := 'my-super-secret-password';
+  BEGIN
+    EXECUTE format('CREATE ROLE %s_%s_writeonly_fivetran;', project, environment);
```
swiknaba, line 14:
```suggestion
    IF NOT EXISTS (SELECT 1 FROM pg_roles WHERE rolname = format('%s_%s_writeonly_fivetran', project, environment))
THEN
    EXECUTE format('CREATE ROLE %s_%s_writeonly_fivetran;', project, environment);
END IF;
```

this will make the script idempotent. We will need to re-execute it if we add/remove tables from the list.

swiknaba (changes requested):
Thanks! Overall LGTM, a few points:
* Please move this to the `scripts` module
* Let's make the "fivetran" part of the name a variable, e.g. "appname"
* the privileges are far more than just "write", I'd suggest a role name like `$project_$environment_sync_$appname`
* see comments to make the script idempotent, ensure permissions are revoked when we alter the allowed-tables array, and the role also has read/delete permissions required for the sync
* add a GRANT to create temporary tables (see link below)
* add a GRANT to create new schemas and tables (see link below)
* could you separate the blocks of code and add a comment above it describing what it does?


https://fivetran.com/docs/destinations/faq/limit-fivetran-permissions

puckzxz:
> Thanks! Overall LGTM, a few points:
> 
> * Please move this to the `scripts` module
> * Let's make the "fivetran" part of the name a variable, e.g. "appname"
> * the privileges are far more than just "write", I'd suggest a role name like `$project_$environment_sync_$appname`
> * see comments to make the script idempotent, ensure permissions are revoked when we alter the allowed-tables array, and the role also has read/delete permissions required for the sync
> * add a GRANT to create temporary tables (see link below)
> * add a GRANT to create new schemas and tables (see link below)
> * could you separate the blocks of code and add a comment above it describing what it does?
> 
> https://fivetran.com/docs/destinations/faq/limit-fivetran-permissions

@swiknaba Thank you for the thorough review! I've addressed all comments and applied your suggestions. This should be good to go now. 🤞

script/database-roles/rds-writeonly-role.sql
```diff
@@ -0,0 +1,42 @@
+-- creates a role to allow syncing data from fivetran to a database
+DO
+$$
+  DECLARE
+    table_name     varchar;
```
swiknaba, line 5:
```suggestion
```
nit: in a `foreach .. in ..` loop, the loop variable is implicitly declared, so this should not be necessary

swiknaba (approved):
Great thanks 👍 
✅ idempotency
✅ clearing of (potentially) lingering privileges before granting new ones
✅ seems like the correct minimum set of permissions required for Fivetran to sync data

One nit pick: add the link to the fivetran docs so one can easily look up why certain (scary looking) permissions are granted.

PR #245, opened 2023-08-17
stack/app: add elasticache major and minor version

samkahchiin:
#### Summary
Add elasticache major and minor version to stack/app module

#### Motivation
To create elasticache using stack/app module. Otherwise it will throw error complaining that the version and parameter group is different

### DIFF
--- a/stack/app/variables.tf
+++ b/stack/app/variables.tf
@@ -155,6 +155,24 @@ variable "elasticache_node_type" {
   default = "cache.t3.micro"
 }
 
+variable "elasticache_major_version" {
+  type    = number
+  default = 7
+  validation {
+    condition     = var.elasticache_major_version >= 6 && var.elasticache_major_version <= 7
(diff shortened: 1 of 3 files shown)


stack/app/variables.tf
```diff
@@ -155,14 +155,32 @@ variable "elasticache_node_type" {
...
+
 variable "elasticache_node_count" {
   type    = number
   default = 1
 }
 
 variable "elasticache_parameter_group_name" {
   type    = string
-  default = "default.redis6.x.cluster.on"
+  default = "default.redis7.cluster.on"
```
swiknaba, line 183: should we make this optional then? it's also optional in the core module https://github.com/dbl-works/terraform/blob/v2023.08.08/elasticache/variables.tf#L32-L35 which creates a custom parameter group if none is given https://github.com/dbl-works/terraform/blob/v2023.08.08/elasticache/elasticache.tf#L92

swiknaba (approved):

PR #249, opened 2023-09-06
rds: add max_allocated_storage variables for rds

samkahchiin:
#### Summary

Add max_allocated_storage to turn on the storage autoscaling

#### Motivation

Enable the rds to autoscale itself when it reach its full storage

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) 
 This project does not follow SemVer, since modules are independent of each other; thus, SemVer does not make sense. Changes are grouped per module.
 
 ## Unreleased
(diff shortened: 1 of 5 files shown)


swiknaba (changes requested):
We have to use this feature with great caution; please add a warning to the readme. We had incidents where a fault in the app would write insane amounts of data to the Database, filling up hundreds of GiBs. If we auto-scale, we might detect such a fault way too late. It is too late, once you scale to 500 GiB, because you cannot easily scale beyond 500 GiB with the gp2 storage, meaning any outage will last even longer.

I'd actually vote to not allow autoscaling at all. If you run out of storage, our first action should be to investigate **why** we run out of storage instead of increasing it. Most likely, we have a problem rather than legitimately lots more than a few hundred GiB of data.

PR #250, opened 2023-09-07
Relax health check requirements

swiknaba:
#### Summary

Some apps take longer to boot up. With the current setting, tasks continue to fail because the health check considers them unhealthy before they finished booting.


samkahchiin (approved):
LGTM. 
Should we make these variable and set those value as default?

swiknaba:
> LGTM. Should we make these variable and set those value as default?

Good idea, I would move these to variables when we move the target group into a reuseable module.

PR #258, opened 2023-09-25
Remove NLB

swiknaba:
#### Summary

* Remove NLB
* if `allowlisted_ssh_ips` variable is set, allow SSH access through the ALB

-> Up for discussion. This has minor security implications since we allow SSH traffic on the load balancer that routes traffic to the application. Mitigation: allow SSH only via fixed IP addresses.

#### Motivation

We don't need an NLB in addition to an ALB. This saves us cost and reduces complexity of the infrastructure. 1 NLB (we have 1 per environment) costs about ~17 $/month.


swiknaba:
Deprioritized.

PR #263, opened 2023-10-03
rds: add log_min_duration_statement, log_retention_period variables, log_min_error_statement

samkahchiin:
#### Summary

log_min_duration_statement determine the statement which exceeds x second to be logged.
If we set it to zero, it means that we will log every statement to the logs.
This will be useful for testing and debugging purpose but not good for the performances as writing logs require disk I/O. 
If a large number of queries are being executed, the constant logging can lead to significant I/O operations, competing with the actual database read/write operations, especially if the logs and database data are on the same disk.

### DIFF
--- a/rds/variables.tf
+++ b/rds/variables.tf
@@ -112,3 +112,15 @@ variable "identifier" {
 locals {
   name = var.name != null ? var.name : "${var.project}-${var.environment}${var.regional ? "-${var.region}" : ""}"
 }
+
+variable "log_min_duration_statement" {
+  type        = number
+  default     = 3000
+  description = "Used to log SQL statements that run longer than a specified duration of time (in ms)."
+}
+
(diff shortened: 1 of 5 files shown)


rds/variables.tf
```diff
@@ -112,3 +112,15 @@ variable "identifier" {
 locals {
   name = var.name != null ? var.name : "${var.project}-${var.environment}${var.regional ? "-${var.region}" : ""}"
 }
+
+variable "log_min_duration_statement" {
+  type        = number
+  default     = 3000
```
swiknaba, line 118: default to "-1" https://aws.amazon.com/blogs/database/working-with-rds-and-aurora-postgresql-logs-part-1/

swiknaba (approved):
Great, thanks!

PR #273, opened 2023-11-23
stack/global: New module for deploying resources required only once per project

swiknaba:


stack/global/chatbot.tf
```diff
@@ -0,0 +1,68 @@
...
+          "logs:CreateLogStream",
+          "logs:DescribeLogStreams",
+          "logs:CreateLogGroup",
+          "logs:DescribeLogGroups"
+        ],
+        "Resource" : "arn:aws:logs:*:*:log-group:/aws/chatbot/*"
+      }
+    ]
+  })
+}
```
swiknaba, line 68: Why are these not inside the chatbot module?

stack/global/cloudtrail.tf
```diff
@@ -1,20 +1,22 @@
...
-  cloudtrail_target_bucket_name      = module.log-ingestor.s3_bucket_name
-  cloudtrail_target_bucket_kms_arn   = module.log-ingestor.s3_kms_arn
+  enable_data_cloudtrail             = var.cloudtrail_config.enable_data_cloudtrail
+  s3_bucket_arns_for_data_cloudtrail = var.cloudtrail_config.s3_bucket_arns_for_data_cloudtrail
+  cloudtrail_target_bucket_name      = module.log-ingestor[0].s3_bucket_name
+  cloudtrail_target_bucket_kms_arn   = module.log-ingestor[0].s3_kms_arn
 }
 
 module "log-ingestor" {
+  count  = var.cloudtrail_config == null ? 0 : 1
```
swiknaba, line 19: I think we should have a different boolean for the `log-ingestor`.

Example 1:
Assume we have 2 accounts, then we launch 2 `stack/global` instances. In both we will need `log-producer`, but only in 1 we will need `log-ingestor`. 

Example 2:
Assume all logs are collected in 1 account that only collects logs. In all other accounts, we launch 1 `stack/global` module, and the only thing we need for Cloutrail is to produce logs, but not to ingest.

stack/global/cloudtrail.tf
```diff
@@ -1,20 +1,22 @@
...
   is_organization_trail              = false
   is_multi_region_trail              = true
   enable_management_cloudtrail       = true
-  enable_data_cloudtrail             = var.enable_data_cloudtrail
-  s3_bucket_arns_for_data_cloudtrail = var.s3_bucket_arns_for_data_cloudtrail
-  cloudtrail_target_bucket_name      = module.log-ingestor.s3_bucket_name
-  cloudtrail_target_bucket_kms_arn   = module.log-ingestor.s3_kms_arn
+  enable_data_cloudtrail             = var.cloudtrail_config.enable_data_cloudtrail
+  s3_bucket_arns_for_data_cloudtrail = var.cloudtrail_config.s3_bucket_arns_for_data_cloudtrail
+  cloudtrail_target_bucket_name      = module.log-ingestor[0].s3_bucket_name
```
swiknaba, line 14: What if this is a 2nd account, that only produces logs? The logs are then ingested by another account, meaning we'd have to pass the target bucket name in as a variable?

I think we want to either pass in an existing ingestor account's information, or we create the ingestor here.

stack/global/outputs.tf
```diff
@@ -0,0 +1,7 @@
+output "ip_sets" {
+  value = module.global-accelerator[0].ip_sets
```
swiknaba, line 2: We need to check if we have a global accelerator enabled, if not, then output a empty string. This will fail without a GA.

stack/global/outputs.tf
```diff
@@ -0,0 +1,7 @@
...
+}
+
+output "dns_name" {
+  value = module.global-accelerator[0].dns_name
```
swiknaba, line 6: same here, without GA return empty string

stack/global/s3-private.tf
```diff
@@ -0,0 +1,20 @@
+module "s3-storage" {
```
swiknaba, line 1: Where do we use this? Why do we need it? If we need it, can we name it by its purpose, e.g. `module "central-log-bucket"` or similar?

stack/global/variables.tf
```diff
@@ -0,0 +1,116 @@
...
+    slack_workspace_name = string
+    platform             = optional(string, "ruby-rails")
+    sentry_teams         = optional(list(string), null)
+    frequency            = optional(number, 30)
+  }))
+  default = {}
+}
+
+variable "period_for_billing_alert" {
+  default     = "28800" # 8 hours
```
swiknaba, line 36: NIT: I think once a day is more than enough.

stack/global/variables.tf
```diff
@@ -0,0 +1,116 @@
...
+
+variable "cloudtrail_config" {
+  type = object({
+    s3_bucket_arns_for_data_cloudtrail = optional(list(string), [])
+    enable_data_cloudtrail             = optional(bool, false)
+  })
+  default = null
+}
+
+variable "private_buckets_list" {
```
swiknaba, line 101: What do we need these for?

stack/global/versions.tf
```diff
@@ -0,0 +1,18 @@
+terraform {
+  required_providers {
+    aws = {
+      source                = "hashicorp/aws"
+      version               = ">= 4.0"
+      configuration_aliases = [aws.us-east-1]
```
swiknaba, line 6: why do we need this?

swiknaba (commented):
@samkahchiin Thank you so much! I think the CloudTrail needs another thought.

stack/global/versions.tf
samkahchiin: The billing alarm can only be setup in us-east-1 region.
I should add a comment for this.

stack/global/variables.tf
samkahchiin: I would like to allow the user to decide how many buckets that would like to create in the global modules. eg. terafform state bucket, central log bucket etc.

stack/global/cloudtrail.tf
samkahchiin: > In both we will need log-producer, but only in 1 we will need log-ingestor.

Assume we have 2 accounts, we will only need 
Account A: 1 log-producer (cloudtrail only)
Account B: 1 log-ingester (cloudtrail + s3)
Account C: 1 log-producer (cloudtrail only)

Right?

swiknaba: Correct!

stack/global/s3-private.tf
```diff
@@ -1,4 +1,4 @@
-module "s3-storage" {
```
swiknaba: NIT: Got it now, thanks for your comment; if we also use it for Terraform state, maybe just call it "global-buckets" and leave a comment with examples why we want these?

swiknaba (commented):
Great thanks! "approved" (just can't approve PRs that I opened)

Issue #185, opened 2023-03-24
Change RDS Storage Settings


swiknaba:
New default settings:
* 10 GiB of storage --> this is already a variable; change the default
* storage autoscaling in 5 GiB increments --> introduce as a variable so we can configure this; suggestion: setting increment to 0 GiB means we **do not have** autoscaling
* storage autoscaling limited to a maximum of 250 GiB --> introduce as a variable
* if not yet available: introduce a variable to skip having snapshots and backups (for staging or test DBs --> default: we have snapshots & backups)

Reasoning: saves costs; from what we see we barely need 100 GiB, and also don't use up all that IOPS. Also: Scaling up is no problem, but scaling down is impossible.

PR #274, opened 2023-11-23
rds: update rds default settings

samkahchiin:
#### Summary
https://github.com/dbl-works/terraform/issues/185

### DIFF
--- a/rds/variables.tf
+++ b/rds/variables.tf
@@ -11,12 +11,22 @@ variable "ca_cert_identifier" {
   type    = string
 }
 
+variable "backup_retention_period" {
+  default = 7
+  type    = number
+}
+
 variable "instance_class" { default = "db.t3.micro" }
 variable "engine_version" {
   default = "14"
   type    = string
 }
 variable "allocated_storage" { default = 10 }
+
+variable "max_allocated_storage" {
+  type    = number
+  default = 0
+}
 
 variable "publicly_accessible" {
   type    = bool
--- a/rds/variables.tf
+++ b/rds/variables.tf
@@ -16,7 +16,7 @@ variable "engine_version" {
   default = "14"
   type    = string
 }
-variable "allocated_storage" { default = 100 }
+variable "allocated_storage" { default = 10 }
 
 variable "publicly_accessible" {
   type    = bool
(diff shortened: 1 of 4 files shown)


rds/variables.tf
```diff
@@ -11,12 +11,22 @@ variable "ca_cert_identifier" {
...
+
 variable "instance_class" { default = "db.t3.micro" }
 variable "engine_version" {
   default = "14"
   type    = string
 }
-variable "allocated_storage" { default = 100 }
+variable "allocated_storage" { default = 10 }
+
+variable "max_allocated_storage" {
```
swiknaba, line 26: For clarity: can we name this e.g. `storage_autoscaling_uppler_limit`? "max_allocated_storage" does not really tell me what it stands for (I think this is bad naming from Terraform).

rds/variables.tf
```diff
@@ -11,12 +11,22 @@ variable "ca_cert_identifier" {
...
+  type    = number
+}
+
 variable "instance_class" { default = "db.t3.micro" }
 variable "engine_version" {
   default = "14"
   type    = string
 }
-variable "allocated_storage" { default = 100 }
+variable "allocated_storage" { default = 10 }
```
swiknaba, line 24:
```suggestion
variable "allocated_storage" {
  type    = number
  default = 10
}
```

for consistency.

stack/app/variables.tf
```diff
@@ -297,6 +297,16 @@ variable "rds_subnet_group_name" {
...
 }
 
+variable "rds_backup_retention_period" {
+  default = 7
+  type    = number
+}
+
+variable "rds_max_allocated_storage" {
+  type    = number
+  default = 0
```
swiknaba, line 307: Suggestion: Should we default to auto-scaling up to 20 GiB? +10 GiB won't hurt, and we have a bit of a safe-guard.

swiknaba (changes requested):
Great thanks! Let's please rename "max_allocated_storage" to something that makes it obvious that this is related to storage autoscaling. Terraform chose a terrible name here.

swiknaba (dismissed):
Specs to be fixed, other than that LGTM! Thanks for improving the meaning.

swiknaba (approved):

PR #275, opened 2023-11-29
chore: cleanup and remove redundant variables

samkahchiin:
### ecs
- Remove region from variables.  [#275](https://github.com/dbl-works/terraform/pull/275)

### rds
- Remove account_id and region from variables. [#275](https://github.com/dbl-works/terraform/pull/275)

### vpc
- Remove account_id and region from variables. [#275](https://github.com/dbl-works/terraform/pull/275)

### vpn
- Remove account_id and region from variables. [#275](https://github.com/dbl-works/terraform/pull/275)

### stack/global
- Update default value for period_for_billing_alert & environment_tags_for_taggable_resources [#275](https://github.com/dbl-works/terraform/pull/275)

### stack/core -> stack/region
- Remove iam-policy-for-taggable-resources module from the module and rename the modules to stack/region [#275](https://github.com/dbl-works/terraform/pull/275)

### stack/app
- Remove account_id from variables.  [#275](https://github.com/dbl-works/terraform/pull/275)

#### Motivation
Clean up and refactor code


rds/iam.tf
```diff
@@ -1,8 +1,11 @@
+data "aws_caller_identity" "current" {}
```
swiknaba, line 1: ⚠️ could this lead to an issue where we apply changes to the _wrong_ account, because we haven't switched the AWS profile locally when we work through multiple projects/accounts?

swiknaba (changes requested):
Generally, it is nice if we need fewer variables. For *region* and *account-id*, I just wonder if that is worth to remove.

My concern is, that if we detect this from the current local profile, we could apply changes to the wrong account.

Do you share this concern, or  are we still safe for a reason I don't see?

rds/iam.tf
samkahchiin: Commonly, we specify the profile directly within the provider block so it is rare to use the wrong profile there.

The issues of the wrong profile frequently happen in the deployment script (eg. `ecs-deploy` folder) which we usually skip the profile setting there. 

Overall, I think the wrong account issues are avoidable in these folders/modules by setting up the profile. Does it assure you? or does it happen to u in the other situation that I don't foresee?

swiknaba: Good thanks! Then fine to clean up everywhere except for the deployment modules. WDYT?

swiknaba (approved):

PR #278, opened 2023-12-05
circleci-token-rotator: new module for rotating circleci secret token

samkahchiin:
#### Summary
new module for rotating circleci secret token

#### Test
Tested in one of the project and verified that the CircleCI flow is still functioning without any authentication issue.


swiknaba (approved):
Great idea! -- nothing blocking!

One suggestion: For the `stack` module, we could simply have `circle_ci_organization_id` as optional input, and if given, we create the token rotator for the `deploy-bot`?

Another suggestion: We create the deploybot here: https://github.com/dbl-works/terraform/blob/v2023.12.22/stack/global/deploy-bot.tf .. should we thus move the rotator to the `stack/global` module?

WDYT? 🤗

swiknaba (dismissed):
Great thanks!

Issue #288, opened 2024-01-16
switch S3 encryption to use default key


swiknaba:
https://github.com/dbl-works/terraform/blob/v2023.12.22/s3-private/s3.tf#L63

We should default to `AES256`, which is the default since January 2023, see https://docs.aws.amazon.com/AmazonS3/latest/userguide/UsingServerSideEncryption.html

Reason: the AWS-provided encryption has no extra costs. When using a self-managed key (even if it's in KMS), AWS charges a fee for each encrpytion/decryption (there is a free quota, so no problem for small projects).

PR #290, opened 2024-01-22
s3-private: set default sse algorithm to AES256

samkahchiin:
#### Summary

- Set sse algorithm to AES256

#### Motivation

https://github.com/dbl-works/terraform/issues/288


s3-private/s3.tf
```diff
@@ -7,8 +7,8 @@ module "s3" {
 
   versioning                  = var.versioning
   kms_deletion_window_in_days = var.kms_deletion_window_in_days
-  enable_encryption           = true
   multi_region_kms_key        = var.multi_region_kms_key
+  enable_encryption           = var.sse_algorithm == "aws:kms"
```
swiknaba, line 11: Should this be `var.sse_algorithm != null`? Why would we not keep enforcing this at "true"?

samkahchiin: `enable_encryption` determine whether the KMS key is to be created. 
Set it to false here because we don't need KMS key if the sse_algorithm is AES256

swiknaba: Thanks, that makes sense. Maybe `enable_encryption` should be named sth. like `with_custom_kms_key`?

s3-private/variables.tf
```diff
@@ -84,3 +84,8 @@ variable "cors_expose_headers" {
   type    = list(string)
   default = ["ETag"]
 }
+
+variable "sse_algorithm" {
+  type    = string
+  default = "AES256"
```
swiknaba, line 90: Can we validate, that the value is either `AES256` or `aws:kms`? The validation is not only a validation, but also documentation for us :)

swiknaba (dismissed):
Thanks! Left some suggestions for improvement, nothing blocking though :)

s3-private/s3.tf
samkahchiin: Will open an issue for this -> https://github.com/dbl-works/terraform/issues/295

swiknaba (approved):

PR #291, opened 2024-01-22
stack/app: set default rds_engine_version to 16

samkahchiin:
#### Summary

Set rds_engine_version default value to 16


swiknaba (approved):
🚀

PR #292, opened 2024-01-22
rds: set nullable to false for engine_version

samkahchiin:
#### Summary

- Set nullable to false for engine_version in rds module

#### Motivation

- We don't need to manage default value in both module


swiknaba (approved):
Great, thanks!

PR #300, opened 2024-02-02
s3-shared: S3 bucket for sharing with guest

swiknaba:


s3-shared/guest-iam.tf
```diff
@@ -0,0 +1,123 @@
...
+  }
+}
+
+resource "aws_iam_user_policy" "s3" {
+  name   = "s3-access-for-${module.s3.bucket_name}"
+  user   = aws_iam_user.guest.name
+  policy = data.aws_iam_policy_document.s3.json
+}
+
+resource "aws_iam_policy" "iam-humans-usage" {
```
swiknaba, line 44: Can we use the module we have for this?

s3-shared/guest-iam.tf
```diff
@@ -0,0 +1,123 @@
...
+      "s3:GetObjectTagging",
+      "s3:GetObjectVersion",
+      "s3:ListBucket",
+      "s3:ListBucketVersions",
+      "s3:PutBucketVersioning",
+      "s3:PutObject",
+      "s3:PutObjectAcl",
+      "s3:PutObjectTagging",
+      "s3:DeleteObject",
+      "s3:DeleteObjectVersion"
```
swiknaba, line 29:
```suggestion
      "s3:GetBucketAcl",
      "s3:GetBucketLocation",
      "s3:GetObject",
      "s3:GetObjectAcl",
      "s3:GetObjectTagging",
      "s3:GetObjectVersion",
      "s3:ListBucket",
      "s3:PutObject",
      "s3:PutObjectAcl",
      "s3:PutObjectTagging",
      "s3:DeleteObject"
```

* we should not allow guests to change the bucket versioning setting
* we should disallow guests to view/change versions of files; this allows us to see if they deleted anything

PR #305, opened 2024-02-11
Chore/fix health check options

swiknaba:


swiknaba (commented):
@samkahchiin found this in pushed branches ... should we contribute this? Did you test using HTTPS for the health check? (to test: web server and one other service, e.g. the pdf-renderer)

samkahchiin:
@swiknaba tested this and the health check will fail if i update it to https :(

PR #314, opened 2024-02-28
azure/container-registry: new module

samkahchiin:

### DIFF
--- a/azure/container-registry/main.tf
+++ b/azure/container-registry/main.tf
@@ -1,0 +1,36 @@
+resource "azurerm_container_registry" "main" {
+  name                          = var.project
+  resource_group_name           = var.resource_group_name
+  location                      = var.region
+  sku                           = var.sku
+  public_network_access_enabled = false
+  anonymous_pull_enabled        = false
+
+  retention_policy {
+    days    = var.retention_in_days
+    enabled = true
+  }
+
+  identity {
+    type         = "UserAssigned"
+    identity_ids = var.user_assigned_identity_ids
+  }
+
+  encryption {
+    enabled            = true
+    key_vault_key_id   = data.azurerm_key_vault_key.main.id
+    identity_client_id = var.encryption_client_id
+  }
+
+  tags = local.default_tags
+}
+
+data "azurerm_key_vault" "main" {
+  name                = local.name
+  resource_group_name = var.resource_group_name
+}
+
+data "azurerm_key_vault_key" "main" {
+  name         = local.name
+  key_vault_id = data.azurerm_key_vault.main.id
+}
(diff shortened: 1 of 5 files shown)


azure/container-registry/variables.tf
```diff
@@ -0,0 +1,75 @@
...
+}
+
+variable "environment" {
+  type = string
+}
+
+variable "admin_enabled" {
+  type    = bool
+  default = true
+}
```
swiknaba, line 24: As discussed, this is fine for now. Can you please add an issue to the backlog to come back to this and refactor this to use either  a service principal or a managed identity? Thanks!

azure/container-registry/variables.tf
```diff
@@ -0,0 +1,75 @@
...
+
+variable "public_network_access_enabled" {
+  type    = bool
+  default = true
+}
```
swiknaba, line 29: same here: good for now, but let's add an issue to disable public access, and instead use a private endpoint.

azure/container-registry/variables.tf
```diff
@@ -0,0 +1,75 @@
...
+  type        = number
+  default     = 7
+  description = "The number of days to retain an untagged manifest after which it gets purged."
+  nullable    = false
+}
+
+variable "sku" {
+  type        = string
+  description = "https://learn.microsoft.com/en-us/azure/container-registry/container-registry-skus"
+  default     = "Basic"
```
swiknaba, line 60: should we default to `"Standard"` to force the user to make a conscious decision for a less secure setup?

azure/container-registry/main.tf
```diff
@@ -0,0 +1,49 @@
...
+      enabled            = true
+      key_vault_key_id   = var.key_vault_key_id
+      identity_client_id = var.encryption_client_id
+    }
+  }
+
+  tags = local.default_tags
+}
+
+# TODO: Research on the lifecycle policy
```
swiknaba, line 37: great point! Please add an issue for this.

swiknaba (dismissed):
LGTM! 🚀

azure/container-registry/variables.tf
```diff
@@ -0,0 +1,75 @@
...
+}
+
+variable "admin_enabled" {
+  type    = bool
+  default = true
+}
+
+variable "public_network_access_enabled" {
+  type    = bool
+  default = true
```
samkahchiin, line 28:
```suggestion
  default = false
```

azure/container-registry/variables.tf
samkahchiin, line 60:
```suggestion
  default     = "Standard"
```

PR #316, opened 2024-02-28
azure/database/postgres: new module

samkahchiin:

### DIFF
--- a/azure/database/postgres/main.tf
+++ b/azure/database/postgres/main.tf
@@ -1,0 +1,40 @@
+resource "azurerm_postgresql_flexible_server" "main" {
+  name                   = local.name
+  resource_group_name    = var.resource_group_name
+  location               = var.region
+  version                = var.version
+  delegated_subnet_id    = azurerm_subnet.main.id
+  private_dns_zone_id    = azurerm_private_dns_zone.main.id
+  administrator_login    = var.administrator_login
+  administrator_password = var.administrator_password
+  create_mode            = var.create_mode
+  storage_mb             = var.storage_mb
+  storage_tier           = var.storage_tier
+  sku_name               = var.sku_name
+
+  identity {
+    type         = "UserAssigned"
+    identity_ids = var.user_assigned_identity_ids
+  }
+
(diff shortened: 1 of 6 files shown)


azure/database/postgres/variables.tf
```diff
@@ -0,0 +1,118 @@
+variable "name" {
```
swiknaba, line 1: should we add a setting for `high_availability`? See https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/postgresql_flexible_server#high_availability

azure/database/postgres/main.tf
```diff
@@ -0,0 +1,46 @@
...
+  version                = var.postgres_version
+  delegated_subnet_id    = azurerm_subnet.main.id
+  private_dns_zone_id    = azurerm_private_dns_zone.main.id
+  administrator_login    = var.administrator_login
+  administrator_password = var.administrator_password
+  create_mode            = var.create_mode
+  storage_mb             = var.storage_mb
+  storage_tier           = var.storage_tier
+  sku_name               = var.sku_name
+
```
swiknaba, line 14: should we explicitly set [backup_retention_days](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/postgresql_flexible_server#backup_retention_days) to `7` to ensure we do have this configured? (future versions of the TF module might change this)

azure/database/postgres/main.tf
```diff
@@ -0,0 +1,46 @@
+resource "azurerm_postgresql_flexible_server" "main" {
```
swiknaba, line 1: sadly, it seems terraform does **not** expose the `public_network_access` setting. Let's output [public_network_access_enabled](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/postgresql_flexible_server#public_network_access_enabled) and print that so we can check if this is (hopefully) disabled by default.

swiknaba (dismissed):
Generally LGTM, we can improve on reliability and security I think.

PR #319, opened 2024-03-04
azure/observability: new module

samkahchiin:


azure/observability/monitors/variables.tf
```diff
@@ -0,0 +1,47 @@
...
+}
+
+variable "container_app_id" {
+  type = string
+}
+
+variable "logs_retention_in_days" {
+  type     = number
+  nullable = false
+  default  = 14
```
swiknaba, line 38: We should set this to 90 days for better audibility.

swiknaba (dismissed):
This looks great! Please adjust the default retention period for logs though to be more inline with standards like ISO27001.

azure/observability/monitors/variables.tf
swiknaba, line 38:
```suggestion
  default  = 90
```

PR #320, opened 2024-03-04
azure/blob-storage: new module

samkahchiin:

### DIFF
--- a/azure/blob-storage/variables.tf
+++ b/azure/blob-storage/variables.tf
@@ -30,7 +30,8 @@ variable "lifecycle_rules" {
 }
 
 variable "user_assigned_identity_ids" {
-  type = list(string)
+  type    = list(string)
+  default = []
 }
 
 # BlobStorage: for storing unstructured object data, such as text or binary data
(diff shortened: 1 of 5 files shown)


azure/blob-storage/variables.tf
```diff
@@ -0,0 +1,155 @@
...
+    change_feed_enabled           = optional(bool, true)
+    change_feed_retention_in_days = optional(number, 14)
+    last_access_time_enabled      = optional(bool, true)
+  })
+  default = {}
+}
+
+variable "public_network_access_enabled" {
+  type    = bool
+  default = true
```
swiknaba, line 90: Do we in most cases need public access? Consider defaulting to `false` here, so people make a conscious decision when enabling it.

azure/blob-storage/variables.tf
```diff
@@ -0,0 +1,155 @@
...
+
+# https://learn.microsoft.com/en-us/azure/storage/common/storage-redundancy
+variable "account_replication_type" {
+  type = string
+  # LRS: Least expensive. copies your data synchronously three times within a single physical location in the primary region
+  # ZRS: copies your data synchronously across three Azure AZs in the primary region.
+  # # Secondary Region
+  # GRS: LRS (Primary Region) + LRS in secondary region.
+  # GZRS: ZRS (Primary Region) + LRS in secondary region
+  default = "LRS"
```
swiknaba, line 70: `GZRS` costs roughly twice as much as `LRS`, however, if we store e.g. 100 GiB, the cost difference is 4€ vs. 2€, which is minimal.

I thus suggest we go with `GZRS` as default value for higher reliability.

swiknaba (approved):
Overall, this looks great to me. I left two considerations regarding some default values; let me know what you think.

azure/blob-storage/variables.tf
samkahchiin, line 90:
```suggestion
  default = false
```

azure/blob-storage/variables.tf
samkahchiin, line 70:
```suggestion
  default = "GZRS"
```

PR #327, opened 2024-03-09
chore(azure): allow passing custom names and tags to all resources

swiknaba:
#### Summary

Allow passing custom names & tags to all Azure modules.

Tags & names always have a default name following the convention of this repo (`${project}-${environment}`).

### DIFF
--- a/azure/blob-storage/main.tf
+++ b/azure/blob-storage/main.tf
@@ -55,7 +55,7 @@ resource "azurerm_storage_account" "main" {
     }
   }
 
-  tags = local.default_tags
+  tags = coalesce(var.tags, local.default_tags)
 }
 
 resource "azurerm_storage_container" "main" {
--- a/azure/container-registry/main.tf
+++ b/azure/container-registry/main.tf
@@ -1,5 +1,5 @@
 resource "azurerm_container_registry" "main" {
-  name                          = var.name == null ? var.project : var.name
+  name                          = coalesce(var.name, local.default_name)
   resource_group_name           = var.resource_group_name
   location                      = var.region
   sku                           = var.sku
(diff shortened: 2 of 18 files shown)


azure/blob-storage/main.tf
```diff
@@ -55,7 +55,7 @@ resource "azurerm_storage_account" "main" {
     }
   }
 
-  tags = local.default_tags
+  tags = coalesce(var.tags, local.default_tags)
```
samkahchiin, line 58: Just an idea, WDYT if we use `merge` to keep the default_tags if it is not override?

swiknaba: I'm conflicted. What if users want to use **exactly** their custom tags, and won't expect any other tags to be added?

samkahchiin: Good point.

azure/container-registry/variables.tf
```diff
@@ -66,8 +62,20 @@ variable "sku" {
...
+
+variable "name" {
+  type        = string
+  default     = null
+  description = "Defaults to 'project-environment'."
+}
+
 locals {
-  name = "${var.project}-${var.environment}"
+  default_name = "${var.project}-${var.environment}"
```
samkahchiin, line 77: Container registry name cannot contain special cahracter

samkahchiin (changes requested):

azure/container-registry/main.tf
```diff
@@ -1,5 +1,5 @@
 resource "azurerm_container_registry" "main" {
-  name                          = var.name == null ? var.project : var.name
+  name                          = coalesce(var.name, local.default_name)
```
samkahchiin, line 2: where do we get the default_name local here?

samkahchiin (changes requested):

samkahchiin (approved):

PR #350, opened 2024-06-17
aws/elasticache: allow parameter_group_name to be null

samkahchiin:
#### Motivation
aws_elasticache_parameter_group will only be created if parameter_group_name is null

See [here](https://github.com/dbl-works/terraform/blob/main/aws/elasticache/elasticache.tf#L92)

### DIFF
--- a/aws/elasticache/variables.tf
+++ b/aws/elasticache/variables.tf
@@ -30,8 +30,9 @@ variable "node_count" {
 }
 
 variable "parameter_group_name" {
-  type    = string
-  default = null # "default.redis7.cluster.on"
+  type        = string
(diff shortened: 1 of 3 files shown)


aws/elasticache/variables.tf
```diff
@@ -31,7 +31,7 @@ variable "node_count" {
 
 variable "parameter_group_name" {
   type    = string
-  default = null
+  default = null # "default.redis7.cluster.on"
```
swiknaba, line 34: nit: we could write the default in a `description` tag.

swiknaba (approved):

PR #353, opened 2024-06-27
aws/ecs: allow alb subnet to be defined

samkahchiin:
#### Summary

- Allow alb subnet to be defined in the ecs module
- Allow alb subnet type to be defined in the stack module

#### Motivation
allow the alb to be placed in the private subnet so its endpoint will not be accessible on internet

### DIFF
--- a/aws/ecs/variables.tf
+++ b/aws/ecs/variables.tf
@@ -45,9 +45,6 @@ variable "allow_alb_traffic_to_ports" {
   nullable = false
 }
 
-# Public subnets are where forwarders run, such as a bastion, NAT or proxy
-variable "subnet_public_ids" { type = list(string) }
-
 # Allow containers to access the following resources from inside the cluster
 variable "secrets_arns" {
   type    = list(string)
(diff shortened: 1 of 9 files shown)


aws/ecs/variables.tf
```diff
@@ -215,13 +212,14 @@ variable "enable_container_insights" {
   nullable = false
 }
 
-variable "multi_az" {
-  type    = bool
-  default = false
+variable "alb_subnet_ids" {
+  type = list(string)
+  validation {
+    condition     = length(var.alb_subnet_ids) >= 2
```
swiknaba, line 218: Do we really need more than one subnet? What if we want to deploy e.g. a staging environment to only a single AZ to safe cost?

aws/ecs/variables.tf
```diff
@@ -215,13 +212,14 @@ variable "enable_container_insights" {
...
+    error_message = "At least two subnets in two different Availability Zones must be specified"
```
swiknaba, line 219:
```suggestion
    error_message = "At least two subnets in two different Availability Zones must be specified (AWS limitation)."
```

swiknaba (dismissed):

PR #356, opened 2024-07-03
feat: extract WAF into its own module

swiknaba:

### DIFF
--- a/aws/waf/waf.tf
+++ b/aws/waf/waf.tf
@@ -95,28 +95,48 @@ resource "aws_wafv2_web_acl" "main" {
         allow {}
       }
 
-      statement {
-        or_statement {
-          dynamic "statement" {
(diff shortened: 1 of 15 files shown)


aws/waf/waf.tf
```diff
@@ -104,32 +88,39 @@ resource "aws_wafv2_web_acl" "alb" {
...
         allow {}
       }
 
       statement {
-        byte_match_statement {
-          search_string = var.domain_name
-          field_to_match {
-            single_header {
-              name = "host"
+        or_statement {
```
swiknaba, line 99: alternative solution: exact match on domain_name OR ENDS_WITH on ".domain_name"

samkahchiin (approved):
LGTM.
We should also add this logic to the stack/app module

samkahchiin (approved):

PR #357, opened 2024-07-04
aws/ecs: take out s3 policy for alb access logs

samkahchiin:
#### Summary

Other terraform implementation will overwrite the Bucket policy, so it would be nicer if we take it out.


swiknaba:
we cherry picked this into here https://github.com/dbl-works/terraform/pull/356

PR #359, opened 2024-07-08
aws/iam/iam-policy-for-taggable-resources

samkahchiin:
#### Summary

Add the list of resources for describe/list access in iam

### DIFF
--- a/aws/iam/iam-policy-for-taggable-resources/locals.tf
+++ b/aws/iam/iam-policy-for-taggable-resources/locals.tf
@@ -11,7 +11,10 @@ locals {
     "elasticloadbalancing",
     "elasticache",
     "rds",
-    "s3"
+    "s3",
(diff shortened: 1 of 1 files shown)


aws/iam/iam-policy-for-taggable-resources/locals.tf
```diff
@@ -11,7 +11,10 @@ locals {
     "elasticloadbalancing",
     "elasticache",
     "rds",
-    "s3"
+    "s3",
+    "waf",
+    "wafv2",
+    "waf-regional"
```
swiknaba, line 17:
```suggestion
    "waf-regional",
```

nit: lowers risk that we update the list with a syntax error. Also simpler diffs when extending the list.

swiknaba (dismissed):

PR #373, opened 2025-04-15
Expose variable `storage_type` in RDS module

fguillen:
#### Summary

- Expose the variable `storage_type` in the `rds` module

#### Motivation

We want to change our rds `storage_type` to `gp3` and we need the variable to be exposed


swiknaba (approved):
Cool thanks!

PR #379, opened 2025-07-23
Fix aurora cluster instance resource

morozRed:
#### Summary

When applying changes from terraform using `id` results in it returning ARN which then is throwing an error during apply:

```
operation error RDS: ModifyDBCluster, https response error StatusCode: 400, RequestID: f22e96cd-b89d-4d1d-9a36-c72a90c02ed7, api error InvalidParameterValue: Invalid database cluster identifier:  arn:aws:rds:eu-central-1:0000000000:cluster:some-cluster-identifier
```

#### Motivation

Fix apply action from terraform


swiknaba (approved):
Thank you!

